PR #29, opened 2021-03-26
Modifications for tsql support

b-per:
Hi!

I have been working on a package to port this package to T-SQL. 
Some of the macros require the `adapter.dispatch` call for me to able to override the default version.

If those changes are accepted I should be able to release `dbt-expectations-tsql` shortly.

Cheers,

### DIFF
--- a/macros/schema_tests/_generalized/equal_expression.sql
+++ b/macros/schema_tests/_generalized/equal_expression.sql
@@ -1,4 +1,8 @@
-{%- macro get_select(model, expression, row_condition, group_by) %}
+{% macro get_select(model, expression, row_condition, group_by) -%}
+    {{ adapter.dispatch('get_select', packages = dbt_expectations._get_namespaces()) (model, expression, row_condition, group_by) }}
+{%- endmacro %}
+
+{%- macro default__get_select(model, expression, row_condition, group_by) %}
     select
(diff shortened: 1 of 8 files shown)


clausherther (approved):
LGTM!
Good idea to make those dispatch macros anyway. I'll probably have to go through and also do that for the rest of the tests to be more clean across the project, but will probably hold off until dbt 0.20 when some of the testing stuff changes in a bigger way (which will require rewrites).

clausherther:
@b-per you should see the new 0.2.8 on dbt-hub later today, or you can develop against `main` for now.
(same for your PR against dbt-date, which is now at 0.2.4)

b-per:
Thanks @clausherther !

I have a bit of work to do on the README and the installation instructions but I have a first version of the T-SQL one here: https://github.com/b-per/dbt-expectations-tsql

I realised that tsql-utils and `spark-utils` don't mention `dbt-utils` in the root `packages.yml`, but they also use git submodules from `dbt-utils`. Do you know why this is the case? I have done the same but don't really understand the benefits.

I'll also need to think about how what mechanisms can be put in place to keep `dbt-expectations-tsql` with `dbt-expectations` when macros/integratio tests/integration models change in your repo.

PR #37, opened 2021-04-09
Add expect_table_column_count_to_equal_other_table_times_factor

nysthee:
Hi

We have a use case where we have a table a and we want to validate that table b always has a row count that is equal to the row count in table a times a predefined factor.

### DIFF
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -117,9 +117,6 @@ models:
             max_value: 4
         - dbt_expectations.expect_table_row_count_to_equal_other_table:
             compare_model: ref("data_test")
-        - dbt_expectations.expect_table_row_count_to_equal_other_table_times_factor:
-            compare_model: ref("data_test")
-            factor: 1
         - dbt_expectations.expect_table_column_count_to_equal:
             value: 7
         - dbt_expectations.expect_table_column_count_to_be_between:
@@ -243,3 +240,9 @@ models:
       - name: col_null
         tests:
           - dbt_expectations.expect_column_values_to_be_null
+
+  - name : data_test_factored
+    tests :
+        - dbt_expectations.expect_table_row_count_to_equal_other_table_times_factor:
+            compare_model: ref("data_test")
+            factor: 2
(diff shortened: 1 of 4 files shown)


integration_tests/models/schema_tests/schema.yml
```diff
@@ -117,6 +117,9 @@ models:
             max_value: 4
         - dbt_expectations.expect_table_row_count_to_equal_other_table:
             compare_model: ref("data_test")
+        - dbt_expectations.expect_table_row_count_to_equal_other_table_times_factor:
+            compare_model: ref("data_test")
+            factor: 1
```
clausherther, line 122: Should we maybe introduce a new test column that is factored up something other than `1`?

clausherther (approved):
Hi @nysthee, thanks for submitting this, looks great. I added a small comment re: the integration, but I'll leave it up to you if you want to change that. Let me know either way and I'll go ahead and merge once you're ready.

nysthee:
I agree, makes sense to make the test a bit more meaningful. I will do it and ping you once it’s there!

- - -
Thomas Nys
+32 475 269 452
https://calendly.com/thomasnys

On 9 April 2021 at 16:10:08, Claus Herther ***@***.***) wrote:

@clausherther approved this pull request.

Hi @nysthee, thanks for submitting this, looks great. I added a small comment re: the integration, but I'll leave it up to you if you want to change that. Let me know either way and I'll go ahead and merge once you're ready.

In integration_tests/models/schema_tests/schema.yml:

> @@ -117,6 +117,9 @@ models:
             max_value: 4
         - dbt_expectations.expect_table_row_count_to_equal_other_table:
             compare_model: ref("data_test")
+        - dbt_expectations.expect_table_row_count_to_equal_other_table_times_factor:
+            compare_model: ref("data_test")
+            factor: 1
Should we maybe introduce a new test column that is factored up something other than 1?

nysthee:
hi @clausherther 

I added a new model with factor 2 number of rows of the original table (didn't bother with the content since we only care about the number of rows)

clausherther:
@nysthee looks good, thanks!

clausherther:
I will cut a new release this weekend, might wrap some other work into that. Thanks again for your contribution!

PR #53, opened 2021-04-27
make integration tests TSQL compatible

swanderz:
I'm working on making a shim package for dbt-expectations inside of tsql-utils (https://github.com/dbt-msft/tsql-utils/pull/38).

TSQL doesn't  support nested CTEs, the solution is to break calls to `dbt_utils.date_spine()` out to their own models that are referenced in the first CTE of the downstream model.

I tested these changes and the models at least run, I don't expect any tests to fail.

### DIFF
--- a/integration_tests/models/schema_tests/series_10.sql
+++ b/integration_tests/models/schema_tests/series_10.sql
@@ -1,0 +1,1 @@
+{{ dbt_utils.generate_series(upper_bound=10) }}
--- a/integration_tests/models/schema_tests/series_4.sql
+++ b/integration_tests/models/schema_tests/series_4.sql
@@ -1,0 +1,1 @@
+{{ dbt_utils.generate_series(upper_bound=4) }}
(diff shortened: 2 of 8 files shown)


clausherther (approved):
Looks good! I snuck in a minor change to the end date for the hourly model (this was probably incorrect to begin with) since we're changing this anyway. Thanks!

clausherther:
@swanderz do you need me to cut a release for this, or is having the changes on `main` sufficient for now? Trying to space the releases a bit more if they don't affect the base tests.

swanderz:
> do you need me to cut a release for this, or is having the changes on `main` sufficient for now?

@clausherther all good for now! We're [importing dbt-expecations as a submodule](https://github.com/dbt-msft/tsql-utils/blob/0f5d46ce50a426f76513d74a7b491c04006b4e40/.gitmodules) I wrote more about [the idea here](https://github.com/dbt-msft/tsql-utils/issues/36#issuecomment-809123275).  to importing the `dbt-expectations` macros. The gist is that if users want to use dbt-utils on SQL Server, they can install `tsql-utils` and get started, without having to also install dbt-expectations (though the user will have `sqlserver__` shims of `dbt-expectations` macros installed)

clausherther:
@swanderz very cool!

PR #61, opened 2021-05-11
First spike at getting tests to pass on dbt 0.20.b1

### LINKED ISSUE #59: Update expectations to at least pass integrations tests for dbt==0.20.0-b1
We need to rewrite expectations to return failing rows, not counts

clausherther:
Closes #59

### DIFF
--- a/macros/schema_tests/table_shape/expect_table_column_count_to_equal.sql
+++ b/macros/schema_tests/table_shape/expect_table_column_count_to_equal.sql
@@ -1,6 +1,8 @@
 {%- macro test_expect_table_column_count_to_equal(model, value) -%}
 {%- if execute -%}
 {%- set number_actual_columns = (adapter.get_columns_in_relation(model) | length) -%}
-select {{ (number_actual_columns - value) | abs }}
+select 1
+from (select 1) a
+where {{ number_actual_columns }} != {{ value }}
 {%- endif -%}
 {%- endmacro -%}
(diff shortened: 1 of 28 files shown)


macros/schema_tests/table_shape/expect_table_column_count_to_equal.sql
```diff
@@ -1,6 +1,8 @@
 {%- macro test_expect_table_column_count_to_equal(model, value) -%}
 {%- if execute -%}
 {%- set number_actual_columns = (adapter.get_columns_in_relation(model) | length) -%}
-select {{ (number_actual_columns - value) | abs }}
+select 1
+from (select 1) a
+where {{ number_actual_columns }} != {{ value }}
```
jtcohen6, line 6: This one, and the ones below it, feel tricky—especially to find the right boundary between Jinja calculation and SQL calculation. I could see wanting to return a row that contains two columns, `number_actual_columns` and `value`, and then set `fail_calc` to be their difference.
```sql
{{ config(fail_calc = "abs(number_actual_columns - value)") }}
select {{ number_actual_columns }} as number_actual_columns, {{ value }} as value
```
That way the debug query / stored-in-database table is just slightly more informative. What do you think?

jtcohen6 (commented):
Glad this refactor wasn't too tricky! I'm going to open one other issue, which relates to `dispatch` changes in 0.20.0

macros/schema_tests/table_shape/expect_table_column_count_to_equal.sql
clausherther: Thanks @jtcohen6! I took some of these suggestions into a new branch!

clausherther:
Closing in favor of #82

PR #64, opened 2021-05-11
dispactching expect_table_row_count_to_equal

alieus:
dispactching `expect_table_row_count_to_equal` to create a shim to make it compatible with TSQL

### DIFF
--- a/macros/schema_tests/table_shape/expect_table_row_count_to_equal.sql
+++ b/macros/schema_tests/table_shape/expect_table_row_count_to_equal.sql
@@ -1,4 +1,4 @@
-{%- macro test_expect_table_row_count_to_equal(model,value,row_condition=None) -%}
+{%- macro test_expect_table_row_count_to_equal(model, value, row_condition) -%}
     {{ adapter.dispatch('test_expect_table_row_count_to_equal', packages=dbt_expectations._get_namespaces()) (model,value,row_condition=None) }}
 {% endmacro %}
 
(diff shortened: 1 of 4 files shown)


clausherther (changes requested):
Thanks for adding that, looks good! One small change: could we add spaces after commas in here, like so:
```
(model, value, row_condition=None)
```
Apologies if this isn't consistent elsewhere.

clausherther (changes requested):
Sorry, actually, one more change:
We should not pass None in the dispatch, the main method already does this.
```
(model, value, row_condition)
```

alieus:
> Thanks for adding that, looks good! One small change: could we add spaces after commas in here, like so:
> 
> ```
> (model, value, row_condition=None)
> ```
> 
> Apologies if this isn't consistent elsewhere.

clausherther:
Hi @alieus! Anything else you need from me to make those changes?

alieus:
Hey @clausherther. I believe I made the changes you requested. This should be good to merge. Let me know if you think otherwise.

macros/schema_tests/table_shape/expect_table_row_count_to_equal.sql
```diff
@@ -1,4 +1,4 @@
-{%- macro test_expect_table_row_count_to_equal(model,value,row_condition=None) -%}
+{%- macro test_expect_table_row_count_to_equal(model, value, row_condition) -%}
     {{ adapter.dispatch('test_expect_table_row_count_to_equal', packages=dbt_expectations._get_namespaces()) (model,value,row_condition=None) }}
```
clausherther, line 2: Hey! Sorry, small thing: could we add spaces here as well like you did above?
Also, please keep the `None` default in the macro parameters, but take it out of the dispatch call.
```
{%- macro test_expect_table_row_count_to_equal(model, value, row_condition=None) -%}    
    {{ adapter.dispatch('test_expect_table_row_count_to_equal', packages=dbt_expectations._get_namespaces()) (model, value, row_condition) }}
```

clausherther (changes requested):

clausherther (approved):
LGTM, thanks!

PR #66, opened 2021-05-19
Add expect_column_stdev_to_be_greater_than test

MartinGuindon:
This PR adds a new test to assert whether the stdev of a column is greater than a defined value.

### DIFF
--- a/macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
+++ b/macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
@@ -1,0 +1,42 @@
+{% macro test_expect_column_stdev_to_be_greater_than(model, column_name,
+                                                    greater_than_value,
+                                                    group_by=None,
+                                                    row_condition=None
+                                                    ) -%}
+    {{ adapter.dispatch('test_expect_column_stdev_to_be_greater_than', 
+        packages = dbt_expectations._get_namespaces()) (model, column_name,
+                                                    greater_than_value,
+                                                    group_by,
+                                                    row_condition
+                                                    ) }}
+{%- endmacro %}
+
+{% macro default__test_expect_column_stdev_to_be_greater_than(model, column_name,
+                                                    greater_than_value,
+                                                    group_by,
+                                                    row_condition
+                                                 ) %}
+with calc as (
+    select distinct
+        {% if group_by %}
+            {{ group_by | join(', ') }},
+            stddev({{ column_name }}) over (
+                partition by {{ group_by | join(', ') }}
+            )
+        {% else %}
+            stddev({{ column_name }})
+        {% endif -%}
+
+        as {{ column_name }}_stddev
+
+    from {{ model }}
+
+),
+
+validation_errors as (
+    select * from calc where {{ column_name }}_stddev <= {{ greater_than_value }}
+)
+
+select count(*) from validation_errors
+    
+{% endmacro %}
--- a/macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
+++ b/macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
@@ -19,7 +19,7 @@
                                                               row_condition) %}
 
 with calc as (
-    select distinct
+    select
         {% if group_by %}
             {{ group_by | join(', ') }},
         {% endif -%}
--- a/macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
+++ b/macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
@@ -1,40 +1,40 @@
-{% macro test_expect_column_stdev_to_be_greater_than(model, column_name,
-                                                    greater_than_value,
-                                                    group_by=None,
-                                                    row_condition=None
-                                                    ) -%}
+{% macro test_expect_column_stdev_to_be_greater_than(model, 
+                                                     column_name,
+                                                     value,
+                                                     group_by=None,
+                                                     row_condition=None) -%}
+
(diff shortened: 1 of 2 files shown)


macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
```diff
@@ -0,0 +1,42 @@
...
+                                                    ) }}
+{%- endmacro %}
+
+{% macro default__test_expect_column_stdev_to_be_greater_than(model, column_name,
+                                                    greater_than_value,
+                                                    group_by,
+                                                    row_condition
+                                                 ) %}
+with calc as (
+    select distinct
```
clausherther, line 20: Would this be functionally equivalent to a simple `group by` if specified and a non-window `stdev` function? I'd prefer that to make this a bit more standard SQL. I haven't run the tests yet, but a windowed stdev is even new to something like Snowflake. Not sure this exists everywhere, and it also makes this a little harder to follow than a more standard implementation.

macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
```diff
@@ -0,0 +1,42 @@
+{% macro test_expect_column_stdev_to_be_greater_than(model, column_name,
+                                                    greater_than_value,
```
clausherther, line 2: We've been just using `value` as the parameter name in other macros (I think that came from the GE API). Probably good for now to try to keep these consistent.

macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
```diff
@@ -0,0 +1,42 @@
...
+                                                    group_by=None,
+                                                    row_condition=None
+                                                    ) -%}
+    {{ adapter.dispatch('test_expect_column_stdev_to_be_greater_than', 
+        packages = dbt_expectations._get_namespaces()) (model, column_name,
+                                                    greater_than_value,
```
clausherther, line 8: Very minor, but could we align the parameters on the subsequent lines with or indented with the line above?
e.g.
```python
model,                                                                
column_name,                                                                
value,
...
```

clausherther (changes requested):
Thanks for submitting this! I had a couple of minor things and a question on implementation. I have to also think a little bit about the use cases and maybe circle back re: examples for the docs. -Claus

macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
MartinGuindon: You're absolutely right. I've derived this test from a more complex query. I've refactored the test to remove the window partition.

macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
MartinGuindon: Got it.

macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
MartinGuindon: Done. This was actually a copy-paste from the test_expect_column_sum_to_be_between macro :)

clausherther: Ha, I'm not surprised it's not consistent. I need to lint that, i.e. manually go through it...

macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
```diff
@@ -0,0 +1,42 @@
...
+{%- endmacro %}
+
+{% macro default__test_expect_column_stdev_to_be_greater_than(model, 
+                                                              column_name,
+                                                              value,
+                                                              group_by,
+                                                              row_condition) %}
+
+with calc as (
+    select distinct
```
clausherther, line 22: I don't think we need that `distinct` anymore?

clausherther (changes requested):
I think code-wise this looks good (except that one extra `distinct`), but we'll need to add an integration test to the test project's `schema.yml` [here](https://github.com/calogica/dbt-expectations/blob/main/integration_tests/models/schema_tests/schema.yml). 
If there is not a column already suitable for this, we'll need to add either a test model (I've been avoiding/putting off doing seeds for each test and instead have been generating test data in `select` statements) or a new column to an existing model so I can test this across platforms. Does that make sense?

macros/schema_tests/aggregate_functions/expect_column_stdev_to_be_greater_than.sql
MartinGuindon: Good catch.

MartinGuindon:
> I think code-wise this looks good (except that one extra `distinct`), but we'll need to add an integration test to the test project's `schema.yml` [here](https://github.com/calogica/dbt-expectations/blob/main/integration_tests/models/schema_tests/schema.yml).
> If there is not a column already suitable for this, we'll need to add either a test model (I've been avoiding/putting off doing seeds for each test and instead have been generating test data in `select` statements) or a new column to an existing model so I can test this across platforms. Does that make sense?

Got it. Will work on that tomorrow :)

clausherther:
@MartinGuindon I just pushed the 0.3.4 [release](https://github.com/calogica/dbt-expectations/releases/tag/0.3.4) which includes support for the type of test you're doing here. I *think* these new options should take of what you needed in this PR, so I'll go ahead and mark it closed. Feel free to open this again though if you think we still need it.

```yaml
dbt_expectations.expect_column_stdev_to_be_between:
    min_value: 0
    strictly: True
```

MartinGuindon:
That's awesome and should work nicely! Thank you!

PR #71, opened 2021-05-28
Add support for group_by to expression tests

clausherther:
This PR adds support for `group_by` to any tests using the `expression_is_true` and `expression_between` macros.

### DIFF
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -281,6 +281,14 @@ models:
 
   - name : timeseries_data_grouped
     tests :
+        - dbt_expectations.expect_table_row_count_to_be_between:
+            min_value: 10
+        - dbt_expectations.expect_table_row_count_to_be_between:
+            min_value: 10
+            group_by: [date_day]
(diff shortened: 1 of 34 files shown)


clausherther:
@swanderz just wanted to flag this for you, not sure if you have any shim'd versions of these macros that can use this or should also support `group_by` on the tsql-utils side. 
Some of the integration tests will now pass in a `group_by` parameter, so you might see some failures against main on your end once this is merged...

clausherther:
@swanderz 
eg.

```
    name : timeseries_data_grouped
    tests :
        - dbt_expectations.expect_table_row_count_to_be_between:
            min_value: 10
            group_by: [date_day]
        - dbt_expectations.expect_table_row_count_to_be_between:
            max_value: 10000
            group_by: [group_id]
```

swanderz:
@clausherther  thanks for the heads up. @alieus is almost done w the shim. Just two tests weren't working. To be clear this just adds tests, rights?

clausherther:
@swanderz no, this adds an option parameter to many tests, `group_by`.  This parameter is only used once in an integration test, but if you're running the tests and your shim does not support this parameter, you'll probably get a test failure.

PR #84, opened 2021-06-14
Update description of type check tests

noel:
Updated the descriptions for the column type check so it is more clear that they check the data type of the column, not the values within it.


clausherther (approved):
Love it, thanks @noel !

Issue #90, opened 2021-07-07
expect_column_values_to_be_in_set doesn't seem to be working


ahmedrad:
Hi, 

I'm looking at the source code for `expect_column_values_to_be_in_set` and it doesn't look like it's validating the fields at all.

```
validation_errors as (
    -- values from the model that are not in the set
    select
        v.value_field
    from
        all_values v
        left join
        set_values s on v.value_field = s.value_field
    where
        v.value_field is null
)
```

I think to validate for errors either `v.value` needs to be `s.value` or the `left join` needs to be a `right join` .. I can submit a pull request this week but thought I'd let you know first as it might affect a lot of tests.

clausherther:
Hi @ahmedrad, thanks for submitting this, really good find! You're right, we should make this:
```sql
where
    s.value_field is null
```
I think because of the way this is written, the integration test here unfortunately always passes, so this must have snuck by (good example of why we may have to explicitly test failures in the future).

Happy to accept your PR if you have the time, or I can fix it this week.

clausherther:
Actually, looks like this isn't covered by an integration test, yikes! No wonder.
So, we also have to add a test in https://github.com/calogica/dbt-expectations/blob/main/integration_tests/models/schema_tests/schema.yml

ahmedrad:
Pull request submitted: https://github.com/calogica/dbt-expectations/pull/91

PR #91, opened 2021-07-09
Fix value in set validation

ahmedrad:
This is a fix for this bug: https://github.com/calogica/dbt-expectations/issues/90

### DIFF
--- a/integration_tests/models/schema_tests/data_test.sql
+++ b/integration_tests/models/schema_tests/data_test.sql
@@ -1,11 +1,11 @@
 select
     1 as idx,
     '2020-10-21' as date_col,
-    cast(0 as{{ dbt_utils.type_float() }}) as col_numeric_a,
-    cast(1 as{{ dbt_utils.type_float() }}) as col_numeric_b,
+    cast(0 as {{ dbt_utils.type_float() }}) as col_numeric_a,
+    cast(1 as {{ dbt_utils.type_float() }}) as col_numeric_b,
     'a' as col_string_a,
     'b' as col_string_b,
-    cast(null as{{ dbt_utils.type_string() }}) as col_null
+    cast(null as {{ dbt_utils.type_string() }}) as col_null
 
 union all
 
--- a/macros/math/rand.sql
+++ b/macros/math/rand.sql
@@ -25,3 +25,9 @@
     random()
 
 {%- endmacro -%}
+
+{% macro redshift__rand() %}
+
+    random()
+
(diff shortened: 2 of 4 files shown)


integration_tests/models/schema_tests/data_test.sql
```diff
@@ -1,11 +1,11 @@
 select
     1 as idx,
     '2020-10-21' as date_col,
-    cast(0 as{{ dbt_utils.type_float() }}) as col_numeric_a,
-    cast(1 as{{ dbt_utils.type_float() }}) as col_numeric_b,
+    cast(0 as {{ dbt_utils.type_float() }}) as col_numeric_a,
```
clausherther, line 4: Good catch, not sure how that happened!

macros/math/rand.sql
```diff
@@ -25,3 +25,9 @@
     random()
 
 {%- endmacro -%}
+
+{% macro redshift__rand() %}
+
+    random()
```
clausherther, line 31: Did you mean to include this in this PR? Or was this necessary for you to pass the tests?

integration_tests/models/schema_tests/data_test.sql
clausherther: Interestingly, the old version compiles to 
```sql
cast(0 as
    float64
) as col_numeric_a
```
so shouldn't throw an error but still a good change.

macros/math/rand.sql
ahmedrad: The default function `rand` doesn't exist in redshift so it was throwing an error. I use dbt with redshift that's why I ran into this issue!

integration_tests/models/schema_tests/data_test.sql
ahmedrad: for me it was compiling as `cast(0 asfloat64) as col_numneric_a` so it was throwing an error

macros/math/rand.sql
clausherther: Gotcha, yeah this package doesn't support Redshift very well because I don't have a way to test there. 
I think what I'll do is take this out and move this to a separate PR, since this isn't directly related to the issue you opened. That way, I'll also better remember to add this to the dbt 0.20 branch.

clausherther (approved):
LGTM!

clausherther:
@ahmedrad this and the Redshift `random` PR is now in release 0.3.7, should be on dbt package hub by end of today hopefully. Thanks for your contribution!

Issue #99, opened 2021-07-21
expect_table_columns_to_match_ordered_list is failing on Redshift


mirosval:
After the upgrade to `0.4` the [expectation](https://github.com/calogica/dbt-expectations/blob/ad9767114ff4159fdaafe00064525ce2b50ef743/macros/schema_tests/table_shape/expect_table_columns_to_match_ordered_list.sql) is failing on Redshift with

```
Invalid operation: failed to find conversion function from "unknown" to text;
```

I think this is because redshift cant guess the type of the `ordered_column_names` and `ordered_relation_column_names`. Locally I managed to get it to run by adding `::text` after.

```sql
select
  count(*) as failures,
  count(*) != 0 as should_warn,
  count(*) != 0 as should_error
from (
       with test_data as (

         select
           'COL_A, COL_B'::text as ordered_column_names,
           'COL_A, COL_B'::text as ordered_relation_column_names

       )
       select *
       from test_data
       where ordered_column_names != ordered_relation_column_names

     ) dbt_internal_test
```

clausherther:
Thanks for raising this issue @mirosval ! Interesting, I'll check into this as much as I can, as well as the other issues you've opened. We don't officially support Redshift because we don't have a way to test against a RS cluster. But maybe there are some generic things we can do help with types etc that would also work on Redshift. I'll take a look by end of the week.

Issue #97, opened 2021-07-21
Regex expectations broken on Redshift


mirosval:
Hello, I've upgraded to dbt `0.20.0` and dbt-expectations `0.4`. After that I started seeing the following error:

```Invalid operation: function regexp_matches(character varying, "unknown") does not exist```

My understanding of what happened is that this https://github.com/calogica/dbt-expectations/issues/39 was fixed, which replaced the `regexp_instr` with `regex_matches` on Postgres. However because Redshift falls back to Postgress version of the macro under the new dispatch rules, this is also `regex_matches` is invoked on Redshift. But redshift does not have that function (even though it does have the original regexp_instr): https://docs.aws.amazon.com/redshift/latest/dg/REGEXP_INSTR.html

**Comment by clausherther** [comment]
Interesting, so I had thought that Redshift would use the `default_` implementation, which [matches](https://github.com/calogica/dbt-expectations/blob/ad9767114ff4159fdaafe00064525ce2b50ef743/macros/regex/regexp_instr.sql#L10) the Redshift function. But you're saying dbt uses the Postgres version in this case? Like `Redshift > Postrges > Default`?

**Comment by mirosval** [comment]
> Plugin maintainers: For adapters that inherit from other adapters (e.g. dbt-postgres → dbt-redshift), adapter.dispatch() will now include parent macro implementations as viable candidates (#2923, #3296)

This is from https://github.com/dbt-labs/dbt/releases/tag/v0.20.0 so I think this is what is happening here and why Redshift doesn't use the default.

**Comment by mirosval** [comment]
I've simply added this locally to `dbt_expectations/macros/regex/regexp_instr.sql`:

```jinja
{% macro redshift__regexp_instr(source_value, regexp, position, occurrence) %}
regexp_instr({{ source_value }}, '{{ regexp }}', {{ position }}, {{ occurrence }})
{% endmacro %}
```
(its just duplicated and renamed the `default` adapter to `redshift`) and it seems to work. I can make a PR if you like

PR #101, opened 2021-07-22
Fix types in expect_table_columns_to_match_* tests 

### LINKED ISSUE #98: Expect table columns to match set fails on Redshift
After the upgrade to `0.4` [this](https://github.com/calogica/dbt-expectations/blob/ad9767114ff4159fdaafe00064525ce2b50ef743/macros/schema_tests/table_shape/expect_table_columns_to_match_set.sql) expectation fails with 

```Invalid operation: failed to find conversion function from "unknown" to text; ```

I think its because of the quotes around the `number_columns`and `number_matching_columns`, those were plain numbers before, as far as I can tell.

clausherther:
This hopefully fixes #98  and #99 

* For #98 this remove `''` that were inadvertently added during an earlier PR
* For #99 this explicitly casts the list of columns to a platform specific string type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # dbt-expectations v0.4.0
 
+## Fixes
+* `expect_table_columns_to_match_list` remove `''` to leave columns as numbers ([#98](https://github.com/calogica/dbt-expectations/issues/98))
+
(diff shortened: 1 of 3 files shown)


clausherther:
Hi @mirosval! I think this addresses your 2 issues. It passes for me on all 3 supported platforms (BQ, Snowflake, Postgres). Would you be able to test this PR on Redshift before I merge this in?

mirosval:
@clausherther thank you very much. This appears to fix the issues (apart from https://github.com/calogica/dbt-expectations/issues/97 obviously).

PR #102, opened 2021-07-23
Fix regexp_instr on Redshift

### LINKED ISSUE #97: Regex expectations broken on Redshift

mirosval:
with the change to adapter resolution in dbt 0.20, redshift has been
using the postgres version of the macro, which fails because the
function `regexp_matches` exists on Postgres but not Redshift. While the
`regexp_instr` does exist on Redshift, so the default was correct all
along.

This commit simply adds an explicit adapter for Redshift that is the
same as the default adapter.

This should fix #97

### DIFF
--- a/macros/regex/regexp_instr.sql
+++ b/macros/regex/regexp_instr.sql
@@ -10,6 +10,9 @@
 regexp_instr({{ source_value }}, '{{ regexp }}', {{ position }}, {{ occurrence }})
 {% endmacro %}
 
+{% macro redshift__regexp_instr(source_value, regexp, position, occurrence) %}
+regexp_instr({{ source_value }}, '{{ regexp }}', {{ position }}, {{ occurrence }})
+{% endmacro %}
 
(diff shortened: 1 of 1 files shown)


clausherther:
Very cool, thanks! I'll take a look today and should be able to merge this and #101 this weekend and cut a new release.

clausherther (approved):
LGTM!

Issue #100, opened 2021-07-22
clarification on the use of expect_table_columns_to_match_set


s-vinay:
Data warehouse: Bigquery

The test "expect_table_columns_to_match_set" is comparing the columns list provided in the "column_list" parameter with the columns in the model. When we have some additional column in the list which does not exist in the model the test is still successful.

This design misguides developers / testers in one particular scenario outlined below.
- developer has missed to add one or more column(s), which was required as per requirement, into a model
- as per the requirement, the test set included this column(s) which was missed by the developer
- now when the test is executed, the test does not fail as all other columns in the set exist in the model

Ideally, the test should compare every column provided in the parameter and even if one is missing it should fail. The table column count test will reveal the count differences but this test also should cover this scenario. Thoughts ?

clausherther:
Hi @s-vinay, thanks for raising this issue! This makes sense, the current test is pretty rudimentary and just comparing columns counts. Definitely something we could improve! Is this something you're interested in working on?

s-vinay:
Hello @clausherther, Thanks for the quick comment and I am really surprised how quick you respond to the issues along with your other daily routines.

I can give this a try, I am new to open source contribution and also probably just few weeks old using DBT but not new for SQL.
So, this comes with some challenges but am interested to start this and will come back soon.

Thanks,
Vinay

clausherther:
@s-vinay no worries! If you'd like to contribute, happy to help! But if it's a bit much to start off with, also happy to take this on next week.

s-vinay:
@clausherther, I have tried some changes - please review and your comments will be helpful

PR #103, opened 2021-07-24
Update expect_table_columns_to_match_set.sql - enhancement (issue #100)

s-vinay:
Enhancement proposal against issue calogica#100.

Reason for change:
- when input columns list contains everything in model + additional columns
- the test is successful as per existing design but it could be that the additional columns are needed in the model as per the requirement

Change introduced:
- the input list length is added to the test_data select
- the final select has one more condition to see if input length is higher than the model length

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ models: # or seeds:
 
 Expect the number of columns in a model to be between two values.
 
-*Applies to:* Column
+*Applies to:* Model, Seed, Source
 
 ```yaml
 tests:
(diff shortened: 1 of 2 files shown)


clausherther:
Hi @s-vinay , sorry for sitting on this so long. I think what I want to do is here is refactor this test completely to not rely on list lengths/counts at all. I have this on my list for the next couple of weeks, so hopefully that'll help close #100 and also make the code easier to follow.

clausherther:
Closing in favor of #108

PR #106, opened 2021-07-30
[Feature] Add row_condition on expect_grouped_row_values_to_have_recent_data test

edbizarro:
This PR add the possibility to apply conditions (`row_condition`) to `expect_grouped_row_values_to_have_recent_data` test

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,14 +187,14 @@ models: # or seeds:
             timestamp_column: date_day
             datepart: day
             interval: 1
-            row_condition: 'id is not null' #optional
+            row_condition: "id is not null" #optional
         # or also:
         - dbt_expectations.expect_grouped_row_values_to_have_recent_data:
             group_by: [group_id, other_group_id]
             timestamp_column: date_day
             datepart: day
             interval: 1
-            row_condition: 'id is not null' #optional
+            row_condition: "id is not null" #optional
 ```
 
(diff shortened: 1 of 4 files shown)


clausherther:
@edbizarro thanks so much for this PR, that's a great addition! Not sure why this wasn't included in the first place.
I'll run tests tonight and should be able to merge this weekend.

clausherther:
One potential issue I'm seeing is that if you filter on a particular value for the grouping level,  you may get a false negative. If the `row_condition` results in 0 rows, the test will pass. I think the expectation should be that this fails, since there are 0 recent rows for this condition.
e.g. try this in `schema.yml`

```yaml
- dbt_expectations.expect_grouped_row_values_to_have_recent_data:
    group_by: [group_id]
    timestamp_column: date_day
    datepart: day
    interval: 1
    row_condition: group_id = 5
```
We may have to rethink what the default timestamp is that's returned when there are no matching rows.

clausherther:
Hey @edbizarro ! I updated the `expect_grouped_row_values_to_have_recent_data` test with some extra logic to make sure we don't get false negatives when `row_condition` excludes all rows. The SQL might look a little hacky, but that's the best I could come up with! :) 
I also added `row_condition` to `expect_row_values_to_have_recent_data`. Here I added a default date to use if `row_condition` excludes all rows.
Hope that makes sense, let me know what you think!

edbizarro:
@clausherther that's clever! I tested on my local environment and works great. And i don't think it's look too much hacky, because, you know, we're doing all this in SQL so...

PR #108, opened 2021-08-13
Convert set logic to SQL joins

### LINKED ISSUE #100: clarification on the use of expect_table_columns_to_match_set

clausherther:
Converts set logic to SQL joins in the following tests:
- expect_column_distinct_values_to_equal_set
- expect_column_values_to_be_in_type_list
- expect_table_columns_to_contain_set
- expect_table_columns_to_match_ordered_list
- expect_table_columns_to_match_set

Closes #100

### DIFF
--- a/macros/schema_tests/aggregate_functions/expect_column_distinct_values_to_equal_set.sql
+++ b/macros/schema_tests/aggregate_functions/expect_column_distinct_values_to_equal_set.sql
@@ -7,7 +7,7 @@
 with all_values as (
 
     select distinct
-        {{ column_name }} as value_field
+        {{ column_name }} as column_value
 
     from {{ model }}
     {% if row_condition %}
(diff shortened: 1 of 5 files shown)


clausherther:
Supersedes #103

Issue #109, opened 2021-09-15
add interval argument to expect_row_values_to_have_data_for_every_n_datepart


lewisarmistead:
We manage large amounts of power industry-related time series data and need to validate presence of data at a granularity between minutely and hourly, e.g. we expect many of our tables to have data at every 5-minute interval.

I see that the [`expect_row_values_to_have_data_for_every_n_datepart`](https://github.com/calogica/dbt-expectations/blob/main/macros/schema_tests/distributional/expect_row_values_to_have_data_for_every_n_datepart.sql) test uses the [`get_base_dates`](https://github.com/calogica/dbt-date/blob/main/macros/get_base_dates.sql) macro from dbt-date to create the date spine used to identify missing data; it would be really helpful to have a line in the [`base_dates`](https://github.com/calogica/dbt-expectations/blob/74d0c218a4fbd7f701420d2b2fba3cef15a70147/macros/schema_tests/distributional/expect_row_values_to_have_data_for_every_n_datepart.sql#L42-L46) CTE to filter the spine to return only X-date_part intervals where X is the value of a new argument - something like this where I used `every_n_datepart` as the new arg:

```sql
{% set start_date = '2021-09-14' %}
{% set end_date = '2021-09-15' %}
{% set date_part = 'minute' %}
{% set every_n_datepart = 5 %}

with base_dates as (

    {{dbt_date.get_base_dates(start_date=start_date, end_date=end_date, datepart=date_part)}}
    where mod({{ dbt_date.date_part(date_part, 'date_' + date_part) }}, {{every_n_datepart}}) = 0

),
```

This addition would allow us to check only for presence at the 5-minute interval in this example (at least when I ran it on Snowflake).

The default for this new datepart interval argument could be `None` with a condition for no-op similar to how some of the other arguments are handled:

```sql
with base_dates as (

    {{dbt_date.get_base_dates(start_date=start_date, end_date=end_date, datepart=date_part)}}
    {% if every_n_datepart %}
    where mod({{ dbt_date.date_part(date_part, 'date_' + date_part) }}, {{every_n_datepart}}) = 0
    {% endif %}

),
```

This may better fit as a feature of dbt-date or the dbt_utils.date_spine macro, but this is where we need it most right now.

Regardless of whether this feature gets implemented or not, thanks for the awesome package!

clausherther:
Hi @lewisarmistead! Thanks for this issue, I'll take a look by this weekend. Would you interested in tackling this in a PR?
[...]

PR #110, opened 2021-09-28
Feature/add interval arg to values every n datepart

lewisarmistead:
This PR addresses issue #109 by adding an optional argument for grouping by count of `date_part` in [expect_row_values_to_have_data_for_every_n_datepart](https://github.com/calogica/dbt-expectations/tree/0.4.2#expect_row_values_to_have_data_for_every_n_datepart).

On the note from that issue about whether `mod` will work across databases, I saw that dbt Labs uses `mod` in [one of their cross-DB utility macros](https://github.com/dbt-labs/dbt-utils/blob/b1f52719b4d5e7e96d0ab0b4bcfde18378502abf/macros/cross_db_utils/width_bucket.sql#L15-L18), so I'm assuming it's safe. I verified that the changes run using Snowflake, and I saw that BigQuery, Postgres, Redshift, & Spark all have the same `mod` function as well.

### DIFF
--- a/macros/schema_tests/distributional/expect_row_values_to_have_data_for_every_n_datepart.sql
+++ b/macros/schema_tests/distributional/expect_row_values_to_have_data_for_every_n_datepart.sql
@@ -44,6 +44,17 @@ with base_dates as (
 
     {{ dbt_date.get_base_dates(start_date=start_date, end_date=end_date, datepart=date_part) }}
     {% if interval %}
+    {# 
+        Filter the date spine created above down to the interval granularity using a modulo operation.
+        The number of date_parts after the start_date divided by the integer interval will produce no remainder for the desired intervals, 
+        e.g. for 2-day interval from a starting Jan 1, 2020:
+            params: start_date = '2020-01-01', date_part = 'day', interval = 2
+            date spine created above: [2020-01-01, 2020-01-02, 2020-01-03, 2020-01-04, 2020-01-05, ...]
+            The first parameter to the `mod` function would be the number of days between the start_date and the spine date, i.e. [0, 1, 2, 3, 4 ...]
+            The second parameter to the `mod` function would be the integer interval, i.e. 2
+            This modulo operation produces the following remainders: [0, 1, 0, 1, 0, ...]
+            Filtering the spine only where this remainder == 0 will return a spine with every other day as desired, i.e. [2020-01-01, 2020-01-03, 2020-01-05, ...]
+    #}
     where mod(
             cast({{ dbt_utils.datediff("'" ~ start_date ~ "'", 'date_' ~ date_part, date_part) }} as {{ dbt_utils.type_int() }}),
             cast({{interval}} as {{ dbt_utils.type_int() }})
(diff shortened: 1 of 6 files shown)


clausherther:
@lewisarmistead so looks like at least in Postgres, both arguments to `mod` need to be integers. I added a test to `schema.yml` 
```yaml
  - name: timeseries_data_extended
    tests:
        - dbt_expectations.expect_table_columns_to_match_ordered_list:
            column_list: ["date_day", "row_value", "row_value_log"]
        - dbt_expectations.expect_row_values_to_have_data_for_every_n_datepart:
            date_col: date_day
            date_part: day
        - dbt_expectations.expect_row_values_to_have_data_for_every_n_datepart:
            date_col: date_day
            date_part: day
            interval: 2
```
which fails on me: 
```
Database Error in test dbt_expectations_expect_row_values_to_have_data_for_every_n_datepart_timeseries_data_extended_date_day__day__2 (models/schema_tests/schema.yml)
  function mod(double precision, integer) does not exist
  LINE 157:     where mod(date_part('day', date_day), 2) = 0
                      ^
  HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

clausherther:
@lewisarmistead btw, I'm not sure I super understand the use case (or the implementation of it). I know you're mostly interested in hourly intervals, but something like `day / interval: 7` would simply check whether we have data on the 7th/14th/21nd/28th of each month (i.e. is the day number divisible by 7 without remainder), which doesn't seem to have a lot of use cases as far as I can tell. Based on the updated README I would think that having data for "each  7-day" period implies every nth day which I can see some use for, but I don't think that's what this implementation allows?

clausherther:
Re: the type issue above, this may work (I can't see needing non-integer arguments?)
```
    {% if interval %}
    where mod(
            cast({{ dbt_date.date_part(date_part, 'date_' + date_part) }} as {{ dbt_utils.type_int() }}),
            cast({{interval}} as {{ dbt_utils.type_int() }})
        ) = 0
    {% endif %}
```

lewisarmistead:
@clausherther - apologies for the intermittent attention here, but thanks for the great feedback! I started work on your suggestions and will push up some changes for your review soon. 

RE: integer arguments for the `mod` function - I agree completely with forcing the type to int. Would it be worth error handling for non-int inputs? e.g. if someone uses 0.5 days for the interval?

RE: checking data presence over a period rather than each X interval - I did misunderstand this, and I'm close to getting the proper aggregation developed for the condition when the interval arg is passed. I'll have something for your review soon on that front too.

clausherther:
@lewisarmistead thanks for working on this and contributing in the first place! Let me know if I can help with anything here.

lewisarmistead:
Thanks @clausherther - I added some changes reflecting the comments above. I didn't want to slow down execution of the original test, so I favored no-op conditions over readability/concision. Let me know your thoughts!

clausherther:
Hi @lewisarmistead! Really appreciate you refactoring this, I think even with the section that concats the mod statement, this looks a lot cleaner to me (we had two uses of mod previously I think?)
I don't think, at this point, I'd invest more in refactoring out that section into a macro, since as you point, it'd be pretty narrow in scope. 
However, since that section is pretty dense, could we just add a solid comment on what exactly is going in there? Maybe even with a quick example of what you're calculating there? Without seeing data, I have a hard time unraveling the dateadd/mod/datediff logic.
Other than that, I think we're there!

macros/schema_tests/distributional/expect_row_values_to_have_data_for_every_n_datepart.sql
```diff
@@ -42,12 +43,34 @@
...
 ),
 model_data as (
 
     select
+    {% if not interval %}
+
         cast({{ dbt_utils.date_trunc(date_part, date_col) }} as {{ dbt_expectations.type_datetime() }}) as date_{{ date_part }},
+
+    {% else %}
+
```
clausherther, line 62: Would be great to add comments, along with an example showing sample data, of what this section does.

clausherther (changes requested):
Once we add some comments to the outlined section, I think we're good to go. Thanks!

macros/schema_tests/distributional/expect_row_values_to_have_data_for_every_n_datepart.sql
lewisarmistead: I added some comments but had some difficulty keeping them brief. Please edit as you see fit, and feel free to reach out with any questions

lewisarmistead:
> 
> 
> Once we add some comments to the outlined section, I think we're good to go. Thanks!

Thanks, @clausherther! Let me know if you need anything else

clausherther (approved):
LGTM!

lewisarmistead:
> 
> 
> @lewisarmistead thanks for all the code comments and hanging in there with me as we worked through this!

Happy to contribute - thanks for the feedback @clausherther !

Issue #125, opened 2021-11-16
Error in macros/schema_tests/string_matching/*


samantha-guerriero-cko:
All the methods in macros/schema_tests/string_matching/ define a group_by_columns=group_by functionality, but group_by is not defined in the method which makes them fail with **_'MacroGenerator' object is not iterable_**.

Fix should be to set up group_by=None in methods' definition as per the other macros with this functionality.

e.g.,
[ {% test expect_column_value_lengths_to_equal(model, column_name,
                                                    value,
                                                    group_by=None,
                                                    row_condition=None
                                                    ) %}](url)

clausherther:
Hi @samantha-guerriero-cko , thanks for raising that issue! You're right, I see the missing params. What's weird is that this passes our integration tests without an error. For example, this test gets called [here](https://github.com/calogica/dbt-expectations/blob/5510fc68638d043faaba691bee5f630964761cd6/integration_tests/models/schema_tests/schema.yml#L281).
Could you share some more information about your environment (which platform, what version dbt & Python)?

Also, let me know if you're interested in submitting a PR for this before I jump on this myself.

clausherther:
Btw, I think the correct treatment here is probably to not support `group_by` for this particular test, since the expression is not an aggregation so it would fail on the group by anyway.

So, I'd set the `group_by_columns` param to `None` in the call to `expression_is_true`:

```python
{{ dbt_expectations.expression_is_true(model,
                                        expression=expression,
                                        group_by_columns=None,
                                        row_condition=row_condition
                                        )
                                        }}
```

samantha-guerriero-cko:
> Btw, I think the correct treatment here is probably to not support `group_by` for this particular test, since the expression is not an aggregation so it would fail on the group by anyway.
> 
> So, I'd set the `group_by_columns` param to `None` in the call to `expression_is_true`:
> 
> ```python
> {{ dbt_expectations.expression_is_true(model,
>                                         expression=expression,
>                                         group_by_columns=None,
>                                         row_condition=row_condition
>                                         )
>                                         }}
> ```

It makes sense to make the group_by_columns=None :)

For what concerns my environment:
- dbt = 0.21.0 (have tried also with 0.20.0)
- Python = 3.9.6
- Platform = macOS Big Sur 11.6

I am happy to open a PR for this 💪

clausherther:
@samantha-guerriero-cko amazing, thank you!

PR #126, opened 2021-11-16
Fix missing group_by default value in string_matching macros

samantha-guerriero-cko:
Fix for issue #125.

Problem: All the methods in macros/schema_tests/string_matching/ define a group_by_columns=group_by functionality, but group_by is not defined in the method which makes them fail with 'MacroGenerator' object is not iterable.

Solution: we set group_by = None as it's not used by the method.


clausherther:
@samantha-guerriero-cko thanks for the super quick turnaround here! Will do a release later today, but may take a bit before it shows up on the dbt package hub.

samantha-guerriero-cko:
thank you @clausherther 💪

PR #127, opened 2021-12-03
Upgrade project to support dbt 0.1.x

clausherther:
Updates the following files to support dbt==0.1.x

* .gitignore
* dbt_project.yml
* integration_tests/dbt_project.yml

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 
 target/
-dbt_modules/
 dbt_packages/
 logs/
 .python-version
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 
 target/
-dbt_modules/
+dbt_packages/
 logs/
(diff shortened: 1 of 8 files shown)


clausherther:
cc: @joellabes

joellabes:
[Same comment as on dbt-date](https://github.com/calogica/dbt-date/pull/50#issuecomment-985842603) - looks good, but would love a patch so that people who can’t jump into 1.0 with both feet can still dip their toes in without getting errors if possible!

clausherther:
> [Same comment as on dbt-date](https://github.com/calogica/dbt-date/pull/50#issuecomment-985842603) - looks good, but would love a patch so that people who can’t jump into 1.0 with both feet can still dip their toes in without getting errors if possible!

Makes sense. Let me see what I can do here over the weekend.

PR #136, opened 2022-01-09
Fix documentation to note DBT 1.0.0 compatibility

kdw2126:
The README.md notes that the project only supports dbt 0.21.x, but dot-expectations 0.5.0 and above support 1.0.0.  

Just wanted to quickly change the documentation to minimize confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 ## Install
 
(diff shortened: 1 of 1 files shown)


clausherther (approved):
Good catch! 👍

PR #138, opened 2022-01-11
Create expect_column_set_to_be_unique_case_insensitive.sql

agusfigueroa-htg:
Compares if the unique values of a given column remain unique from a case insensitive perspective. Useful when dealing with CamelCase and lowercase uniqueness within a column. Example: the test would fail if "MasterCard" and "mastercard" are in the same column.

I don't know the guidelines of the commits, but I can take care of adding the corresponding links and description to the README.


clausherther:
Hi @agusfigueroa-htg, thanks for your PR! The use case here makes sense to me. However, I wonder if we could instead modify `expect_column_values_to_be_unique` to take in an optional boolean `case_sensitive` parameter that would accomplish the same thing? 
We don't have strict contribution guidelines (yet), but it'd be great if you could ultimately add to/update the README. Also, if you could, please take a look at the existing macros for formatting guidelines. Thanks!

agusfigueroa-htg:
Hey @clausherther, thanks for reviewing this! Question on the comment, as I gave that some thought but still I don’t fully see it. 

In this test the column values are not unique, I just want the set of unique values to remain unique case insensitive.

Some examples below

Example 1:
“iPhone”
“iPhone”
expect_column_values_to_be_unique  fails
expect_column_set_to_be_unique_case_insensitive  passes

Example 2:
“iPhone”
“iphone”
expect_column_values_to_be_unique  passes
expect_column_set_to_be_unique_case_insensitive  fails

expect_column_set_to_be_unique_case_insensitive doesn’t make much sense in my eyes as a case-sensitive test - because I am already considering the column set of values, which is unique by definition.

**Is there any existing test that comes to your mind to integrate this use case? Otherwise, we can leave it standalone. Let me know what you think.**

Regarding the next steps, I can adjust whatever is needed, format it as the other macros and add the documentation to the readme.

Thanks again! :)

clausherther:
@agusfigueroa-htg what I meant was, we could add a new optional parameter to `expect_column_values_to_be_unique`, `case_sensitive=False` and based on that parameter value, implement the logic needed to make the uniqueness test case insensitive in the `expect_column_values_to_be_unique` test, without needing a standalone test.
We're trying to avoid too many similar standalone tests. Does that make sense?

agusfigueroa-htg:
@clausherther 
That is feasible but would change the idea of the test in itself as it is not analyzing the uniqueness of the column, but analyzing if the set (unique values of the column) remain unique case insensitive.

Our use case, and what I wonder whether it would come in handy for further users, is to apply this to large tables where the column is not necessarily unique but you wanna ensure that there are no duplicates case insensitive. For instance, when picking the OS column of a sessions table I would expect values to show up there repeatedly (e.g. Windows, Android, iOS, Android, iOS, etc). If, for some data problem, we have sessions with different ways of naming "windows" (e.g. Windows, Android, iOS, Android, windows, iOS, etc) I want the test to fail. Values being repeated is expected (which is why expect_column_values_to_be_unique would fail), what this adds on top is the expectation of that set of values remaining unique from a case insensitive perspective.

I pinged you on slack - maybe we can briefly sync.
Thanks,
Agus

macros/schema_tests/column_values_basic/expect_column_set_to_be_unique_case_insensitive.sql
```diff
@@ -0,0 +1,24 @@
+{% test expect_column_set_to_be_unique_case_insensitive(model, column_name) %}
+
+ {%- if execute -%}
```
clausherther, line 3: don't think we need the `if execute`?

macros/schema_tests/column_values_basic/expect_column_set_to_be_unique_case_insensitive.sql
```diff
@@ -0,0 +1,24 @@
...
+ with test_data as (
+
+     select
+         distinct {{ column_name }} as distinct_values
+         FROM
```
clausherther, line 8: we use all lower-case SQL in the repo, and align with the `select` clause

macros/schema_tests/column_values_basic/expect_column_set_to_be_unique_case_insensitive.sql
```diff
@@ -0,0 +1,24 @@
...
+         {{ model }}
+         where 1=1
+
+ ),
+ validation_errors as
+ (
```
clausherther, line 14: just a style point: opening `(` goes with previous line

macros/schema_tests/column_values_basic/expect_column_set_to_be_unique_case_insensitive.sql
```diff
@@ -0,0 +1,24 @@
...
+   select
+   count(1) as set_count,
```
clausherther, line 16: indent please 🙏

agusfigueroa-htg:
After discussing with @clausherther, we aligned on the fact that what we were looking for was a test for **consistent casing**. 
This means, that if a column has several repeated values it's fine, as long as the casing is consistent. 


```
--Example of expected values for a given column: 

google
google
google
aws
aws
othervalue
```

```
--Example of unexpected values for a given column: 

google
**G**oogle
google
aws
aws
othervalue
```

As in this case google is named differently (google and Google), thus the casing is inconsistent. For this reason, we decided to make this a standalone test, and add a flag so the test can return the faulty column values if needed. 

The final PR is here, including formatting and documentation, feel free to review it and get back to me if needed.
Thanks

macros/schema_tests/column_values_basic/expect_column_values_to_have_consistent_casing.sql
```diff
@@ -0,0 +1,36 @@
+{% test expect_column_values_to_have_consistent_casing(model, column_name, display_inconsistent_columns=False) %}
+
+with test_data as (
+
+     select
+     distinct {{ column_name }} as distinct_values
```
clausherther, line 6: Nitpick: please indent the statements under keywords, e.g.
```sql
select
    distinct {{ column_name }} as distinct_values
from
    {{ model }}
```
throughout the macro. Thanks!

macros/schema_tests/column_values_basic/expect_column_values_to_have_consistent_casing.sql
```diff
@@ -0,0 +1,36 @@
...
+ ),
+ {% if display_inconsistent_columns %}
+ validation_errors as (
+
+    select
+    lower(distinct_values) as inconsistent_columns,
+    count(distinct_values) as set_count_case_insensitive
+    from test_data
+    group by 1
+    having count(distinct_values)>1
```
clausherther, line 20: Formatting: we typically do spaces around operator, e.g. `1 = 1` not `1=1`. So, in this case `count(distinct_values) > 1` for better readability.

macros/schema_tests/column_values_basic/expect_column_values_to_have_consistent_casing.sql
```diff
@@ -0,0 +1,36 @@
...
+with test_data as (
+
+     select
+     distinct {{ column_name }} as distinct_values
+     from
+     {{ model }}
+     where 1=1
+
+ ),
+ {% if display_inconsistent_columns %}
```
clausherther, line 12: I'm not sure we need to have two paths here? I think the `display_inconsistent_columns` path is sufficient, no? Any benefit to users of the test in the other case?

clausherther:
@agusfigueroa-htg we'll need to add an integration test [here](https://github.com/calogica/dbt-expectations/blob/28d47f73b198f9ca808aa071630a439859607e99/integration_tests/models/schema_tests/schema.yml) please.

macros/schema_tests/column_values_basic/expect_column_values_to_have_consistent_casing.sql
agusfigueroa-htg: My reasoning was to do with performance. The path that does not display inconsistent columns saves a group by that may be costly when testing huge tables. 

Also, some common tests (like the "column has no nulls") return only one row so I thought doing the same was consistent.

Let me know what you think.

clausherther: Got it, makes sense. Thanks!

clausherther:
Hi @agusfigueroa-htg! Is this PR ready to be reviewed? Let me know if there's anything I can do to help get this over the finish line this week.

agusfigueroa-htg:
Ah yes sorry! When I submitted the last commit I forgot to tag you in a comment. Yes, ready to be reviewed @clausherther . Thanks!

README.md
```diff
@@ -410,6 +411,18 @@ tests:
       column_type_list: [date, datetime]
 ```
 
+### [expect_column_values_to_have_consistent_casing](macros/schema_tests/column_values_basic/expect_column_values_to_have_consistent_casing.sql)
+
+Expect a column to have consistent casing. By setting display_inconsistent_columns to true, the inconsistent columns will be displayed.
```
clausherther, line 416: 1) Small comment: should be `display_inconsistent_columns` (in single markdown backticks)
2) I think we should be more clear here - the columns will only be displayed if you run the compiled test SQL for debugging in your console, nothing will be displayed to the screen during testing right?

integration_tests/models/schema_tests/schema.yml
```diff
@@ -274,9 +274,9 @@ models:
           - dbt_expectations.expect_column_distinct_values_to_contain_set:
               value_set: ['a','b']
               quote_values: true
-
-          - dbt_expectations.expect_column_value_lengths_to_equal   :
+          - dbt_expectations.expect_column_value_lengths_to_equal:
               value: 1
+          - dbt_expectations.expect_column_values_to_have_consistent_casing
```
clausherther, line 279: Can we also please test the `display_inconsistent_columns: true` case since that's very different SQL?

macros/schema_tests/column_values_basic/expect_column_values_to_have_consistent_casing.sql
```diff
@@ -0,0 +1,41 @@
...
+        distinct {{ column_name }} as distinct_values
+    from
+        {{ model }}
+
+ ),
+ {% if display_inconsistent_columns %}
+ validation_errors as (
+
+    select
+        lower(distinct_values) as inconsistent_columns,
```
clausherther, line 15: @agusfigueroa-htg FYI, I updated some of the formatting, so you'll have to pull those changes into your local branch

README.md
agusfigueroa-htg: 1. Fine, will adjust.
2. In the console, the number of rows is gonna reflect the number of inconsistent values in the column. If you run the compiled test, you will get the results.

integration_tests/models/schema_tests/schema.yml
agusfigueroa-htg: Great, added as a new testing instance

agusfigueroa-htg:
@clausherther I introduced those changes, may you check? thx!

clausherther (approved):
LGTM! Thanks!

Issue #104, opened 2021-07-25
expect_row_values_to_have_recent_data fails in BigQuery for TIMESTAMP columns


kdw2126:
If you add a expect_row_values_to_have_recent_data argument to a TIMESTAMP column, you will get a failure driven by two characteristics of the existing implementation:
- [This comparison](https://github.com/calogica/dbt-expectations/blob/e47358d5ee5f162457fd4b40aa9ea84d8879ea93/macros/schema_tests/table_shape/expect_row_values_to_have_recent_data.sql#L8) fails because BigQuery does not allow TIMESTAMP columns to be compared with DATE objects using the <= operator.
- [This comparison](https://github.com/calogica/dbt-expectations/blob/e47358d5ee5f162457fd4b40aa9ea84d8879ea93/macros/schema_tests/table_shape/expect_row_values_to_have_recent_data.sql#L15) fails because BigQuery does not allow TIMESTAMP columns to be compared with DATESTAMP using the <= operator.

Given that a master time zone is specified in vars under dbt_project.yml, it seems possible to rewrite these comparisons using the appropriate recasting statements to allow columns of a TIMESTAMP type to be compatible with the expect_row_values_to_have_recent_data test.

clausherther:
Thanks for this issue @kdw2126 ! This timestamp/date issue has been a problem for a while, I think even `dbt_utils.recency` has this. I'd be happy for any PRs if you have ideas around this, but I'll mull this over next week when I have some spare cycles.

clausherther:
Closed via #145

nirakon:
> Closed via #145

It didn't fix the issue.

`No matching signature for operator < for argument types: TIMESTAMP, DATETIME. Supported signature: ANY < ANY`

clausherther:
@nirakon 😩 it passed tests on BQ, but I'll reopen this and look into it. I think this more of an issue with dbt_date.now() returning a datetime.

bascheibler:
I think it's the `dbt_utils.dateadd` macro that is returning a datetime. It calls the `datetime_add` function from BigQuery, which returns a datetime data type.

So the inner CTE ("max_recency") casts it as timestamp, but the outer CTE ("dbt_internal_test") casts it as datetime.

I'm using the version 0.5.2 with a date and had a different issue: if the column in my dbt model is a DATE, it returns the following error: `No matching signature for operator <= for argument types: DATE, TIMESTAMP. Supported signature: ANY <= ANY`.

[...]

PR #145, opened 2022-02-24
Change check for data date out of range to use timestamp

MrJoosh:
This PR addresses #104 by using timestamps as the comparison to work out older data.

This was driven by experiencing an issue using the test on data using a timestamp as the recency source for data that should be fresher than 6 hours.

The original code was pulling data from the table that had a timestamp <= todays date, meaning that although the most recent data was loaded 15 minutes before, the test was failing.

Tested on Snowflake

Apologies if this doesn't meet the standards for contributions for the repo, but before #139 is implemented, it's justa guess :(

Also please let me know if anything else is required

### DIFF
--- a/macros/schema_tests/table_shape/expect_row_values_to_have_recent_data.sql
+++ b/macros/schema_tests/table_shape/expect_row_values_to_have_recent_data.sql
@@ -20,7 +20,7 @@ with max_recency as (
     from
         {{ model }}
     where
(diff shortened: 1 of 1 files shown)


clausherther (approved):
LGTM! Thanks for the PR!

AdeelK93:
This was a breaking change for me, `expect_row_values_to_have_recent_data` now errors out for me in BQ, and there's no combination of casting that I can do to get it working again

PR #146, opened 2022-02-26
Added partition by in lag tests

Lucasthenoob:
# Problem: 
Not possible to put partition by parameters in the following tests:
- [expect_column_values_to_be_increasing](https://github.com/calogica/dbt-expectations/blob/main/macros/schema_tests/column_values_basic/expect_column_values_to_be_increasing.sql)
- [expect_column_values_to_be_decreasing](https://github.com/calogica/dbt-expectations/blob/main/macros/schema_tests/column_values_basic/expect_column_values_to_be_decreasing.sql)

# Changes:
- Added an optional parameter which allow to partition by. 
- Updated the README.MD with accordingly

Please let me know if there is anything

### DIFF
--- a/README.md
+++ b/README.md
@@ -480,7 +480,7 @@ tests:
       sort_column: date_day
       row_condition: "id is not null" # (Optional)
       strictly: true # (Optional for comparison operator. Default is 'true', and it uses '>'. If set to 'false' it uses '>='.)
-      partition_columns: partition_col_1, partition_col_2 # (Optional: default is None)
+      group_by: [group_id, other_group_id, ...] # (Optional)
 ```
 
 ### [expect_column_values_to_be_decreasing](macros/schema_tests/column_values_basic/expect_column_values_to_be_decreasing.sql)
@@ -497,7 +497,7 @@ tests:
       sort_column: col_numeric_a
       row_condition: "id is not null" # (Optional)
       strictly: true # (Optional for comparison operator. Default is 'true' and it uses '<'. If set to 'false', it uses '<='.)
-      partition_columns: partition_col_1, partition_col_2 # (Optional: default is None)
+      group_by: [group_id, other_group_id, ...] # (Optional)
 ```
 
 ### [expect_column_value_lengths_to_be_between](macros/schema_tests/string_matching/expect_column_value_lengths_to_be_between.sql)
--- a/README.md
+++ b/README.md
@@ -479,14 +479,15 @@ tests:
   - dbt_expectations.expect_column_values_to_be_increasing:
       sort_column: date_day
       row_condition: "id is not null" # (Optional)
-      strictly: true # (Optional for comparison operator. Default is 'true', and it uses '>'. If set to 'flase' it uses '>='.)
+      strictly: true # (Optional for comparison operator. Default is 'true', and it uses '>'. If set to 'false' it uses '>='.)
+      partition_columns: partition_col_1, partition_col_2 # (Optional: default is None)
 ```
 
 ### [expect_column_values_to_be_decreasing](macros/schema_tests/column_values_basic/expect_column_values_to_be_decreasing.sql)
 
 Expect column values to be decreasing.
 
-If strictly=True, then this expectation is only satisfied if each consecutive value is strictly increasing–equal values are treated as failures.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -479,23 +479,25 @@ tests:
   - dbt_expectations.expect_column_values_to_be_increasing:
       sort_column: date_day
       row_condition: "id is not null" # (Optional)
-      strictly: true # (Optional for comparison operator. Default is 'true', and it uses '>'. If set to 'flase' it uses '>='.)
+      strictly: true # (Optional for comparison operator. Default is 'true', and it uses '>'. If set to 'false' it uses '>='.)
+      partition_columns: partition_col_1, partition_col_2 # (Optional: default is None)
```
clausherther, line 483: Let's make `partition_columns` an actual list not a string, similar to https://github.com/calogica/dbt-expectations#expect_table_row_count_to_be_between

README.md
```diff
@@ -479,23 +479,25 @@ tests:
...
 *Applies to:* Column
 
 ```yaml
 tests:
   - dbt_expectations.expect_column_values_to_be_decreasing:
       sort_column: col_numeric_a
-      strictly: true # (Optional for comparison operator. Default is 'true' and it uses '<'. If set to 'false', it uses '<='.)
       row_condition: "id is not null" # (Optional)
+      strictly: true # (Optional for comparison operator. Default is 'true' and it uses '<'. If set to 'false', it uses '<='.)
+      partition_columns: partition_col_1, partition_col_2 # (Optional: default is None)
```
clausherther, line 500: See comment above re: `partition_columns`

macros/schema_tests/column_values_basic/expect_column_values_to_be_decreasing.sql
```diff
@@ -1,16 +1,22 @@
 {% test expect_column_values_to_be_decreasing(model, column_name,
                                                    sort_column=None,
                                                    strictly=True,
-                                                   row_condition=None) %}
+                                                   row_condition=None,
+                                                   partition_columns=None) %}
 
 {%- set sort_column = column_name if not sort_column else sort_column -%}
 {%- set operator = "<" if strictly else "<=" %}
+{%- set partition_columns_list = partition_columns.split(',') %}
```
clausherther, line 9: `partition_columns` should be a Jinja list, so no need to spilt.

macros/schema_tests/column_values_basic/expect_column_values_to_be_decreasing.sql
```diff
@@ -22,7 +28,12 @@ add_lag_values as (
     select
         sort_column,
         value_field,
-        lag(value_field) over(order by sort_column) as value_field_lag
+        lag(value_field) over
+            {%- if not partition_columns -%}
+                (order by sort_column)
+            {%- else -%}
+                (partition by {{ partition_columns }} order by sort_column)
```
clausherther, line 35: Since `partition_columns` is a list, you'll need to either loop, or do a `.join(",")` (easier) here.

macros/schema_tests/column_values_basic/expect_column_values_to_be_increasing.sql
```diff
@@ -1,34 +1,40 @@
 {% test expect_column_values_to_be_increasing(model, column_name,
                                                    sort_column=None,
                                                    strictly=True,
-                                                   row_condition=None) %}
+                                                   row_condition=None,
+                                                   partition_columns=None) %}
 
 {%- set sort_column = column_name if not sort_column else sort_column -%}
 {%- set operator = ">" if strictly else ">=" %}
+{%- set partition_columns_list = partition_columns.split(',') %}
```
clausherther, line 9: Btw, this is never used, so the below doesn't run.

clausherther (changes requested):
Couple of changes re: partition column parameters.
Also, we'll need a test that actually uses these partitions in https://github.com/calogica/dbt-expectations/blob/main/integration_tests/models/schema_tests/schema.yml
I can help you with that if you've never done integration tests on dbt packages. Thanks!

Lucasthenoob:
It should be good now.
Sorry I don't have dbt installed on my personal computer. 
I validated the test is working on my other one but I could not test the second commit: data creation and test implementation

clausherther (approved):
I made a few changes myself to help get this over the finish line. Hope that works for you!

PR #147, opened 2022-03-09
Fix expect_row_values_to_have_recent_data issues on bigquery

### LINKED ISSUE #104: expect_row_values_to_have_recent_data fails in BigQuery for TIMESTAMP columns

clausherther:
This PR attempts to fix #104 by converting all evals in the test macro to force a cast to `timestamp`.

### DIFF
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -24,8 +24,8 @@ models:
 
   - name: timeseries_data
     tests:
-        - dbt_expectations.expect_table_columns_to_match_ordered_list:
-            column_list: ["date_day", "row_value", "row_value_log"]
+        # - dbt_expectations.expect_table_columns_to_match_ordered_list:
+        #     column_list: ["date_day", "row_value", "row_value_log"]
         - dbt_expectations.expect_column_distinct_count_to_equal_other_table:
             column_name: date_day
             compare_model: ref("timeseries_data_extended")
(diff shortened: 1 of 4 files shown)


clausherther:
Hi @bascheibler and @AdeelK93! Could I ask one or both of you to give this branch a shot? This is just a quick spike to see if this might fix the issue. It's a little ugly with all the casts, but datetimes are ugly. Thanks!

clausherther:
Merging this as I think this, at last for now, addresses #104

Issue #148, opened 2022-03-10
feature/request - add a test to show if specific column(s) have been added to an object


rcaddell:
Specific use case is Servicenow data loaded to Snowflake via Fivetran. Fivetran will not add a _fivetran_deleted column to the Servicenow object unless and until at least 1 record from the source object was deleted.

PR #149, opened 2022-03-10
Adds test for column presence

rcaddell:
Adds table shape test to look for the population of certain columns. Test is a copy of the expect_table_columns_to_contain_set with the final where clause changed

Specific use case is Servicenow data loaded via Fivetran. Fivetran will not add a _fivetran_deleted column to the Servicenow object unless and until at least 1 record from the source object was deleted. We've added this test to any Servicenow source that does not already have that column. This allows us to leave Fivetran set to bring in all new columns and only notify us if _fivetran_deleted was added and needs to be addressed in the dbt project.  #148

### DIFF
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -167,7 +167,7 @@ models:
         - dbt_expectations.expect_table_column_count_to_equal_other_table:
             compare_model: ref("data_test")
         - dbt_expectations.expect_table_columns_to_not_contain_set:
-            column_list: ["col_numeric_b", "col_string_a"]
+            column_list: ["col_numeric_c", "col_string_d"]
         - dbt_expectations.expression_is_true:
             expression: (col_numeric_a + col_numeric_b = 1)
         - dbt_expectations.equal_expression:
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -166,6 +166,8 @@ models:
             column_list: ["idx", "date_col", "col_numeric_a", "col_numeric_b", "col_string_a", "col_string_b", "col_null"]
         - dbt_expectations.expect_table_column_count_to_equal_other_table:
             compare_model: ref("data_test")
+        - dbt_expectations.expect_table_columns_to_not_contain_set:
+            column_list: ["col_numeric_b", "col_string_a"]
         - dbt_expectations.expression_is_true:
             expression: (col_numeric_a + col_numeric_b = 1)
         - dbt_expectations.equal_expression:
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -330,3 +330,19 @@ models:
             datepart: day
             interval: 1
             row_condition: group_id = 4
+
+  - name: window_function_test
+    columns:
+      - name: rolling_sum_increasing
+        tests :
+          - dbt_expectations.expect_column_values_to_be_increasing:
+              group_by: ['idx']
+              strictly: true
+              sort_column: date_col
+
+      - name: rolling_sum_decreasing
+        tests :
+          - dbt_expectations.expect_column_values_to_be_decreasing:
+              group_by: ['idx']
+              strictly: true
+              sort_column: date_col
--- a/macros/schema_tests/table_shape/expect_table_columns_not_to_contain_set.sql
+++ b/macros/schema_tests/table_shape/expect_table_columns_not_to_contain_set.sql
@@ -1,0 +1,29 @@
+{%- test expect_table_columns_not_to_contain_set(model, column_list, transform="upper") -%}
+{%- if execute -%}
+    {%- set column_list = column_list | map(transform) | list -%}
+    {%- set relation_column_names = dbt_expectations._get_column_list(model, transform) -%}
+    {%- set matching_columns = dbt_expectations._list_intersect(column_list, relation_column_names) -%}
+    with relation_columns as (
+
+        {% for col_name in relation_column_names %}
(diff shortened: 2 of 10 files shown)


macros/schema_tests/table_shape/expect_table_columns_not_to_contain_set.sql
```diff
@@ -0,0 +1,29 @@
...
+        {% endfor %}
+    )
+    select *
+    from
+        input_columns i
+        left join
+        relation_columns r on r.relation_column = i.input_column
+    where
+        -- catch any column in input list that is not in the list of table columns
+        r.relation_column is not null
```
clausherther, line 27: Doesn't this `where` clause just turns this into an `inner join`? I think we should just rewrite the `left join` as inner join.

macros/schema_tests/table_shape/expect_table_columns_not_to_contain_set.sql
```diff
@@ -0,0 +1,29 @@
...
+        {% if not loop.last %}union all{% endif %}
+        {% endfor %}
+    )
+    select *
+    from
+        input_columns i
+        left join
+        relation_columns r on r.relation_column = i.input_column
+    where
+        -- catch any column in input list that is not in the list of table columns
```
clausherther, line 26: I think the comment should read:
```
-- catch any column in input list that is in the list of table columns
```
Since this section should be refactored into an `inner join`, this comment should also move up.

macros/schema_tests/table_shape/expect_table_columns_not_to_contain_set.sql
```diff
@@ -0,0 +1,29 @@
+{%- test expect_table_columns_not_to_contain_set(model, column_list, transform="upper") -%}
```
clausherther, line 1: To match the other `*_not_*` tests, I think `expect_table_columns_to_not_contain_set` would be a better name. See: `expect_column_values_to_not_be_in_set`

clausherther (changes requested):
Hi @rcaddell , thanks for this PR! I don't super understand the Fivetran use case, but I think the test makes sense and has applicability in general terms.
Couple of small code comments and a name change to get this over the finish line.
Also, we'll need to add a test in `integration_tests/schema.yml`. Let me know if you need help with that and the integration tests in general.

macros/schema_tests/table_shape/expect_table_columns_not_to_contain_set.sql
rcaddell: it does, I will address

macros/schema_tests/table_shape/expect_table_columns_not_to_contain_set.sql
rcaddell: updated

macros/schema_tests/table_shape/expect_table_columns_not_to_contain_set.sql
rcaddell: updated

rcaddell: Thanks @clausherther , I may need help with the integration tests.  There is no data_test.csv file to seed when I run dbt seed from the integration_tests folder

rcaddell:
Thanks @clausherther , I may need help with the integration tests.  There is no data_test.csv file to seed when I run dbt seed from the integration_tests folder

clausherther:
@rcaddell we don't use seeds, all test data comes from the models in the `integration_tests/models` folder. Since there are so many tests, we try to test the test macros on a small number of test models. E.g. a single model with a number of columns can be used to test a large number of test macros, just as you would use them in practice. Have a look at the integration_tests/models folder to see what I mean. But happy to commit an example to your branch once you push the other changes. Thanks!

rcaddell:
Hi @clausherther, I need some help with the integration testing if you could.  I have an 'integration_tests' profile configured with needed access as per the readme.  I added a test to schema.yml following the same pattern as the 'expect_table_columns_to_contain_set' entry, which errors with 2 results.  I also receive an error in building the emails.sql model 

>syntax error line 1 at position 30 unexpected '-' 

Any guidance?  Thanks!

clausherther:
Hi @rcaddell let me check out your PR locally and give it a shot. I'll get back to you here by EOW. Sorry about the delay!

integration_tests/models/schema_tests/schema.yml
```diff
@@ -166,6 +166,8 @@ models:
             column_list: ["idx", "date_col", "col_numeric_a", "col_numeric_b", "col_string_a", "col_string_b", "col_null"]
         - dbt_expectations.expect_table_column_count_to_equal_other_table:
             compare_model: ref("data_test")
+        - dbt_expectations.expect_table_columns_to_not_contain_set:
+            column_list: ["col_numeric_b", "col_string_a"]
```
clausherther, line 170: this should probably be something like
```
column_list: ["col_numeric_c", "col_string_d"]
```
to make the test pass since we're looking for columns *not* in the model?

clausherther (changes requested):
I wasn't able to commit changes to your PR for some reason. Is your fork private?
I fixed the issue with `emails.sql` in `main`, so could you please rebase your branch and try running the integration test again? At a minimum, you'll want a local postgres instance and run `dbt build`. Thanks!

integration_tests/models/schema_tests/schema.yml
rcaddell: Updated and tests passed

rcaddell:
Hi @clausherther, my fork is public from what I can see, possible my enterprise admin has a config set to block?

Branch rebased and test updated based on your feedback.  all tests passed.  

Thanks!

clausherther (approved):
LGTM! Thanks!

PR #152, opened 2022-03-14
fix: quote column_name before passing to dbt_utils.length in expect_column_value_lengths_to_be_between

### LINKED ISSUE #150: expect_column_value_lengths_to_be_between doesn't work with a column name as a reserved keyword
`expect_column_value_lengths_to_be_between` doesn't quote the column name. If the column name is a reserved keyword (such as `offset`, `start`, etc... with pg), the test fails.

tevariou:
Fixes #150

If the column name is a reserved keyword, `expect_column_value_lengths_to_be_between` tests would fail because the column name should be quoted before being passed to `dbt_utils.length`. This PR adds quotes around `column_name` in this generic test.

### DIFF
--- a/integration_tests/models/schema_tests/data_test.sql
+++ b/integration_tests/models/schema_tests/data_test.sql
@@ -5,7 +5,8 @@ select
     cast(1 as {{ dbt_utils.type_float() }}) as col_numeric_b,
     'a' as col_string_a,
     'b' as col_string_b,
-    cast(null as {{ dbt_utils.type_string() }}) as col_null
+    cast(null as {{ dbt_utils.type_string() }}) as col_null,
+    'a reserved keyword' as "offset"
 
 union all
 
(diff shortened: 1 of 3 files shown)


clausherther:
@tevariou thanks for this PR! Actually, I think dbt handles this natively. If you include `quote: true` in your config at the column's `config` level (or maybe even higher up), dbt will automatically quote column names. I pushed a commit to this PR to show what I mean. Check it out, and if this works for you, I'd say we merge this to have in example in our integration tests, but no need to update the tests.

clausherther:
Btw, quoting columns like that, e.g.`"offset"` isn't valid on BigQuery, so we wouldn't be able to merge this since we'd break tests on BQ which is one of our main supported platforms.
On BigQuery, this works, 
```sql
'a reserved keyword' as offset
```

tevariou:
Thanks a lot for your help @clausherther

PR #158, opened 2022-04-02
Show group_by columns in validation errors for column increasing test

dluftspring:
# Description

Quick suggestion - if there are multiple group_by keys it's a lot easier to look for test failures if we include them in the select and then they show up in validation errors.

There's probably a reason i'm not thinking of that this wasn't included initially - will close if that's the case 😄


clausherther (approved):
I don't see a reason why we shouldn't include the group_by columns. Since we're not actually grouping, this shouldn't change the number of output rows.

PR #176, opened 2022-05-19
Support ephemeral models in expectations

judahrand:
Current ephemeral models do not work with some tests provided by `dbt-expectations`. This PR attempts to support them through the use of `adapter.adapter.get_columns_in_query`.

### DIFF
--- a/integration_tests/models/schema_tests/data_test_ephemeral.sql
+++ b/integration_tests/models/schema_tests/data_test_ephemeral.sql
@@ -1,0 +1,47 @@
+{{
+    config(
+        materialized='ephemeral',
+    )
+}}
+
+select
+    1 as idx,
+    '2020-10-21' as date_col,
+    cast(0 as {{ dbt_utils.type_float() }}) as col_numeric_a,
+    cast(1 as {{ dbt_utils.type_float() }}) as col_numeric_b,
+    'a' as col_string_a,
+    'b' as col_string_b,
+    cast(null as {{ dbt_utils.type_string() }}) as col_null
+
+union all
+
+select
+    2 as idx,
+    '2020-10-22' as date_col,
+    1 as col_numeric_a,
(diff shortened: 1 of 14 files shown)


clausherther:
Hey @judahrand thanks for this PR! I'll take a more detail look, but for now looks like the integration tests are failing with error `'obj' is undefined` (at least on postgres), using dbt 1.1.0.

judahrand:
I'll have a look when I get a chance!

judahrand:
@clausherther I think I fixed the error are you able to allow the integration tests to run again?

clausherther:
@judahrand looks like we still got some database errors as well as test failures
https://app.circleci.com/pipelines/github/calogica/dbt-expectations/58/workflows/0d2372ac-5bd1-443a-9d9e-95c8b1ca5c36/jobs/50?invite=true#step-104-370

clausherther:
@judahrand I think the issue is that the ephemeral model still returns `{'type': 'PostgresRelation'}`, ` {'type': 'SnowflakeRelation'}` etc, just like the `table` materialization. 
I think you'll want to use `model.is_cte` in this case, see https://docs.getdbt.com/reference/dbt-classes#using-relations
However, I get this error when dbt is trying to get columns for a CTE:
```
 'dbt.context.providers.RuntimeDatabaseWrapper object' has no attribute 'get_columns_in_query'
```
Are you sure this is supposed to work for ephemeral models/CTEs?

clausherther:
I don't think you can do column introspection on ephemeral models.
See: https://github.com/dbt-labs/dbt-utils/blob/main/macros/cross_db_utils/_is_ephemeral.sql

clausherther:
Hey @judahrand ! Just wanted to check in if this is something you're still pursuing or if we should close this PR? Thanks!

PR #191, opened 2022-09-03
Add option to escape raw strings in RegEx functions

### LINKED ISSUE #190: Tests that expect regex patterns are painful to use in BigQuery
Functions such as [expect_column_values_to_match_regex](https://github.com/calogica/dbt-expectations/blob/main/macros/schema_tests/string_matching/expect_column_values_to_match_regex.sql) are painful to use in BigQuery because of the intersection of YAML escape sequences, BigQuery escape sequences and implementation of the tests, by dbt-expectations (I'm not sure actually if dbt-core has some responsibility here to fix this so do say if the fix doesn't belong here).

Given some trivial regex expression, `"\d{4}"`, in BigQuery, I would make use of raw strings to avoid illegal escape sequences, e.g. `r"\d{4}"` but dbt-expectations/dbt expects my yaml to be a string and so I can't use `r` otherwise it'll be interpreted as `'r"\d{4}"'`.

Probably dbt-expectations should be smart enough to insert r in front of the string if the adapter supports it (such as BigQuery).

In the meantime, I probably have to double escape the `\`. Once to satisfy YAML escaping and twice to satisfy BigQuery escaping when a raw string is not used.

E.g. I probably need to do:
```yaml
    regex: "\\\d{4}" # note that the encasing "" are irrelevant here.
```

clausherther:
Closes #190

This PR adds functionality to allow the RegEx-based tests, such as `expect_column_values_to_match_regex` to accept unescaped/raw string patterns, by adding an optional `is_raw` parameter to the relevant tests. The default is `False`, i.e. `regex` strings are not automatically escaped.

For more context:
BigQuery: uses `r'...'` to mark the string as "raw"
Snowflake:  uses `$$..$$` to avoid needing escape characters (see https://docs.snowflake.com/en/sql-reference/functions-regexp.html#specifying-regular-expressions-in-dollar-quoted-string-constants)


clausherther:
@adamcunnington-mlg would your eyes on this when you get a chance. Thanks!

clausherther:
> @adamcunnington-mlg would love your eyes on this when you get a chance. Thanks!

I'm going to go ahead and merge this for the upcoming release, since it's a non-breaking change.

PR #202, opened 2022-10-01
Fix boolean logic in `expect_compound_columns_to_be_unique` `ignore_row_if` 

### LINKED ISSUE #200: `expect_compound_columns_to_be_unique` has backward logic for `ignore_row_if`
Caveat, it's possible I've just misunderstood the intended semantics. If so, sorry.

## What

I noticed when using [expect_compound_columns_to_be _unique](https://github.com/calogica/dbt-expectations#expect_compound_columns_to_be_unique) that it passed when I expected it to fail, evidently because it was not including records with any nulls in its test. 

The docs read: 
> ignore_row_if: "any_value_is_missing" # (Optional. Default is 'all_values_are_missing')

Which led me to believe all rows would be included so long as _any_ field was non-null. I thought perhaps the default was wrong, but specifying `"all_values_are_missing"` explicitly gave the same result. When I changed to `"any_value_is_missing"`, the test went from passing to failing. 

Inspecting the rendered SQL showed me something like: 

`ignore_row_if: "all_values_are_missing"`
```sql
with validation_errors as (

    select
        g1,g2,g3
    from dbt_cannamm_dev_macro_tests.test_dbt_expectations__expect_compound_columns_to_be_unique
    where
        1=1
        and
        (
            g1 is not null and
            g2 is not null and
            g3 is not null

        )

    group by
        g1,g2,g3
    having count(*) > 1

)
select * from validation_errors
```

You can see that it's selecting where "all fields are not null" for inclusion in the test; but the negation of "all fields are not null" (i.e. the "ignored" rows) is "any field is null", the opposite of what the arguments specified. 

Looking a the [code](https://github.com/calogica/dbt-expectations/blob/91abad064513150f502b6751b1170443dac3bb88/macros/schema_tests/multi-column/expect_compound_columns_to_be_unique.sql#L31) bears this out: 

```
{%- set row_condition_ext -%}


{%- if ignore_row_if == "all_values_are_missing" %}
        (
            {% for column in columns -%}
            {{ column }} is not null{% if not loop.last %} and {% endif %}
            {% endfor %}
        )
{%- elif ignore_row_if == "any_value_is_missing" %}
        (
            {% for column in columns -%}
            {{ column }} is not null{% if not loop.last %} or {% endif %}
            {% endfor %}
        )
{%- endif -%}

...elided...

where
        1=1
    {%- if row_condition_ext %}
        and {{ row_condition_ext }}
    {% endif %}
```

Amusingly, since the `ignore_row_if` argument is not guarded against invalid values like the others, one can access a third behavior ("don't ignore any rows") by passing in any other string. 

## Reproduction

Here are (failing) "unit tests" in dbt that show the behavior: 

`test_dbt_expectations__expect_compound_columns_to_be_unique.sql`
```sql
WITH setup AS (
    SELECT g1, g2, g3
    FROM VALUES
             (null, null, null),
             (null, null, null),
             (null, null, 1),
             (null, null, 1),
             (1, null, 1),
             (null, 'a', 2),
             (1, 'a', 3),
             (1, 'b', 4),
             (2, 'a', 5),
             (2, 'b', 6)
              AS t(g1, g2, g3)
)
select *
from setup
```

`test_dbt_expectations__expect_compound_columns_to_be_unique.yml`
```yml
version: 2

expect_to_fail: &exp_fail
    fail_calc: '(count(*)=0)::int'

models:
  - name: test_dbt_expectations__expect_compound_columns_to_be_unique
    description: >
      This test proves how the dbt_expectations test `expect_compound_columns_to_be_unique` works, since the behavior of
      `ignore_row_if` is unclear.

    tests:
      - dbt_expectations.expect_table_row_count_to_equal:
          value: 9

      - dbt_expectations.expect_compound_columns_to_be_unique:
          column_list: [ 'g1', 'g2' ]
          ignore_row_if: "any_value_is_missing"
          config:
            meta:
              comment: "Passes even in the presence of degeneracy if rows with any null are skipped"


      - dbt_expectations.expect_compound_columns_to_be_unique:
          column_list: [ 'g1', 'g2', 'g3' ]
          ignore_row_if: "all_values_are_missing"
          config:
            <<: *exp_fail
            meta:
              comment: "Fails when rows with some nulls are degenerate in non-null columns and we only skip if all values are null"

      - dbt_expectations.expect_compound_columns_to_be_unique:
          column_list: [ 'g1', 'g2', 'g3' ]
          ignore_row_if: ""
          config:
            <<: *exp_fail
            meta:
              comment: "Fails when rows with some nulls are degenerate in non-null columns and we only skip if all values are null"

      - dbt_expectations.expect_compound_columns_to_be_unique:
          column_list: [ 'g1', 'g2', 'g3' ]
          ignore_row_if: "any_value_is_missing"
          config:
            meta:
              comment: "Passes even in the presence of degeneracy if rows with any null are skipped"

      - dbt_expectations.expect_compound_columns_to_be_unique:
          column_list: [ 'g1', 'g2', 'g3' ]
          config:
            <<: *exp_fail
            meta:
              comment: "Should fail by default if `ignore_row_if` defaults to 'all_values_are_missing' as the docs claim"

```

## Suggested fix
Reversing the bodies of the [if branches](https://github.com/calogica/dbt-expectations/blob/91abad064513150f502b6751b1170443dac3bb88/macros/schema_tests/multi-column/expect_compound_columns_to_be_unique.sql#L31) should give the expected behavior. 

It would probably also be wise to add an explicit "never_ignore" option and guard against users passing nonsense values for `ignore_row_if` that bypass the logic. 

I guess it's worth pointing out that this will be very surprising for folks who have been using this macro and suddenly have a bunch of failing tests, but I don't imagine there is any avoiding that.

clausherther:
Fixes #200 

This fixes the implementation of the `ignore_row_if` parameter in `expect_compound_columns_to_be_unique` highlighted by @mcannamela in #200 and adds a test to `data_test`.

### DIFF
--- a/macros/schema_tests/multi-column/expect_compound_columns_to_be_unique.sql
+++ b/macros/schema_tests/multi-column/expect_compound_columns_to_be_unique.sql
@@ -9,6 +9,13 @@
         "`column_list` must be specified as a list of columns. Got: '" ~ column_list ~"'.'"
     ) }}
 {% endif %}
+{%- set ignore_row_if_values = ["all_values_are_missing", "any_value_is_missing"] -%}
+{% if ignore_row_if not in ignore_row_if_values %}
+    {{ exceptions.raise_compiler_error(
+        "`ignore_row_if` must be one of " ~ (ignore_row_if_values | join(", ")) ~ ". Got: '" ~ ignore_row_if ~"'.'"
+    ) }}
+{% endif %}
+
 {% if not quote_columns %}
     {%- set columns=column_list %}
 {% elif quote_columns %}
--- a/macros/schema_tests/multi-column/expect_compound_columns_to_be_unique.sql
+++ b/macros/schema_tests/multi-column/expect_compound_columns_to_be_unique.sql
@@ -28,19 +28,12 @@
     {{ row_condition }} and
 {% endif -%}
 
-{%- if ignore_row_if == "all_values_are_missing" %}
-        (
-            {% for column in columns -%}
-            {{ column }} is not null{% if not loop.last %} and {% endif %}
-            {% endfor %}
-        )
-{%- elif ignore_row_if == "any_value_is_missing" %}
-        (
-            {% for column in columns -%}
-            {{ column }} is not null{% if not loop.last %} or {% endif %}
-            {% endfor %}
-        )
-{%- endif -%}
+{%- set op = "and" if ignore_row_if == "all_values_are_missing" else "or" -%}
+    not (
+        {% for column in columns -%}
+        {{ column }} is null{% if not loop.last %} {{ op }} {% endif %}
+        {% endfor %}
+    )
 {%- endset -%}
 
 with validation_errors as (
--- a/integration_tests/models/schema_tests/data_test.sql
+++ b/integration_tests/models/schema_tests/data_test.sql
@@ -5,7 +5,8 @@ select
     cast(1 as {{ type_float() }}) as col_numeric_b,
     'a' as col_string_a,
     'b' as col_string_b,
-    cast(null as {{ type_string() }}) as col_null
+    cast(null as {{ type_string() }}) as col_null,
+    cast(null as {{ type_string() }}) as col_null_2
 
 union all
 
@@ -16,7 +17,8 @@ select
     0 as col_numeric_b,
     'b' as col_string_a,
     'ab' as col_string_b,
-    null as col_null
+    null as col_null,
+    null as col_null_2
 
 union all
 
@@ -27,7 +29,8 @@ select
     0.5 as col_numeric_b,
     'c' as col_string_a,
     'abc' as col_string_b,
-    null as col_null
+    null as col_null,
+    null as col_null_2
 
 union all
 
@@ -38,4 +41,5 @@ select
     0.5 as col_numeric_b,
     'c' as col_string_a,
     'abcd' as col_string_b,
-    null as col_null
+    null as col_null,
+    null as col_null_2
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -207,9 +207,9 @@ models:
             ignore_row_if: "all_values_are_missing"
             config:
                 # this should fail, so we flip the fail condition
-                fail_calc: '(count(*)=0)::int'
+                fail_calc: 'cast((count(*)=0) as int)'
         - dbt_expectations.expect_compound_columns_to_be_unique:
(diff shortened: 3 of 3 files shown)


macros/schema_tests/multi-column/expect_compound_columns_to_be_unique.sql
```diff
@@ -28,19 +28,12 @@
...
-            {% endfor %}
-        )
-{%- elif ignore_row_if == "any_value_is_missing" %}
-        (
-            {% for column in columns -%}
-            {{ column }} is not null{% if not loop.last %} or {% endif %}
-            {% endfor %}
-        )
-{%- endif -%}
+{%- set op = "and" if ignore_row_if == "all_values_are_missing" else "or" -%}
```
mcannamela, line 31: This looks correct now, but closes off the "third behavior" of "never ignore" and somewhat obfuscates what behavior to expect if I pass a value other than `"all_values_are_missing"`. 

Did you mean to eliminate the (heretofore undocumented) third behavior?

What would you think of adding a guard prior to this that ensures the value of `ignore_row_if` is an expected one? Even if it's not documented, anybody reading the code would then know how to call it.

clausherther: > I also had the thought, is it possible that this logic shows up in other expectations? I didn't investigate this at all.
I'll take a look, good call out. If so, I'll open a separate issue/PR though.

clausherther:
> I also had the thought, is it possible that this logic shows up in other expectations? I didn't investigate this at all.

I'll take a look, good call out. If so, I'll open a separate issue/PR though.

macros/schema_tests/multi-column/expect_compound_columns_to_be_unique.sql
clausherther: So, the built-in uniqueness check [always checks for nulls](https://github.com/dbt-labs/dbt-core/blob/23c8ac230ccf3e56413d3fa7d22fc62c5e51d7a3/core/dbt/include/global_project/macros/generic_test_sql/not_null.sql#L7) and excludes null values from the uniqueness check.
The default for the multi-column case (which is the generalized version of `expect_column_values_to_be_unique`) is to exclude rows where both columns are null from the uniqueness test, unless overridden.

I think I makes sense to check that the value of `ignore_row_if` is from a known set and error out otherwise.

mcannamela: sounds good, i agree with this default behavior

PR #240, opened 2023-02-08
Escaping single quotes in column names

RobbertDM:
There's probably other places where this applies
I've had this problem with trino, not sure if this applies to different DBs (do other DBs even allow for column names to have single quotes?)
If this needs to be supported, probably a good idea to include some crazy column names in the tests


clausherther:
Thanks @RobbertDM for this PR. I'm not sure I've encountered this issue, not sure how one even goes about adding single quotes in a column name?
Let me think about if this needs to be done anywhere else and if so, we'd need to add a test before merging.
Is this a blocker for you?

clausherther:
Ok, got it. Feel free to open an issue if this continues to be an issue for you.

PR #247, opened 2023-03-10
Add optional flags argument to regex tests

### LINKED ISSUE #245: [Feature Request] Allow regex flags/parameters in regexp_instr and calling tests
I would like the regex I provide to `expect_column_values_to_match_regex_list` and similar tests to be case-insensitive, but currently that is not possible.

[Postgres](https://www.postgresql.org/docs/current/functions-matching.html), [Snowflake](https://docs.snowflake.com/en/sql-reference/functions/regexp_instr), and [Redshift](https://docs.aws.amazon.com/redshift/latest/dg/REGEXP_INSTR.html) support an argument (called `flags` in pg and `parameters` in the others) for this. [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#regexp_instr) does not, but the column could be lowercased before being passed into regexp_instr.

Add a `flags` argument to `expect_column_values_to_match_regex_list` and other regex-based tests that accepts a string value and passes it through to the underlying database implementation.

Rolling my own test.

**Additional context**
I'm happy to contribute this if there is interest.

tconbeer:
Closes #245 

I chose to validate the flag against the lists in the documentation for Postgres, Snowflake, and Redshift. I decided to raise an error if flags are passed to BQ (since they are not supported) or the "default" implementation (unclear support). 

One alternative to raising on all flags for those adapters would be to specifically check for the `i` flag and lowercase the column before passing it to regexp_instr.

### DIFF
--- a/macros/regex/regexp_instr.sql
+++ b/macros/regex/regexp_instr.sql
@@ -27,7 +27,7 @@ regexp_instr({{ source_value }}, {{ regexp }}, {{ position }}, {{ occurrence }},
 {% macro bigquery__regexp_instr(source_value, regexp, position, occurrence, is_raw, flags) %}
 {% if flags %}
     {{ exceptions.raise_compiler_error(
-            "The flag option is not supported for by BigQuery"
+            "The flag option is not supported for BigQuery"
     ) }}
 {% endif %}
 {%- set regexp = "r'" ~ regexp ~ "'" if is_raw else "'" ~ regexp ~ "'" -%}
--- a/macros/regex/regexp_instr.sql
+++ b/macros/regex/regexp_instr.sql
@@ -9,8 +9,9 @@
 {% macro default__regexp_instr(source_value, regexp, position, occurrence, is_raw, flags) %}
 {# unclear if other databases support raw strings or flags #}
 {% if is_raw or flags %}
-    {{ exceptions.raise_compiler_error(
-            "is_raw and flags options not supported for this adapter"
+    {{ exceptions.warn(
+            "is_raw and flags options are not supported for this adapter "
+            ~ "and are being ignored."
     ) }}
 {% endif %}
 regexp_instr({{ source_value }}, '{{ regexp }}', {{ position }}, {{ occurrence }})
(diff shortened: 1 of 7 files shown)


clausherther:
Cool, thanks @tconbeer, will take a look this weekend. 
Btw, the compatibility check for BQ causes CI to fail, so we'll have to come with another way to handle this before this can be merged:
```
The flag option is not supported for by BigQuery
```

macros/regex/regexp_instr.sql
```diff
@@ -1,33 +1,57 @@
...
-regexp_instr({{ source_value }}, {{ regexp }}, {{ position }}, {{ occurrence }})
+{% if flags %}{{ dbt_expectations._validate_flags(flags, 'cimes') }}{% endif %}
+regexp_instr({{ source_value }}, {{ regexp }}, {{ position }}, {{ occurrence }}, 0, '{{ flags }}')
 {% endmacro %}
 
 {# BigQuery uses "r" to escape raw strings #}
-{% macro bigquery__regexp_instr(source_value, regexp, position, occurrence, is_raw) %}
+{% macro bigquery__regexp_instr(source_value, regexp, position, occurrence, is_raw, flags) %}
+{% if flags %}
+    {{ exceptions.raise_compiler_error(
```
clausherther, line 29: We may need to make this a compiler warning?

tconbeer: happy to downgrade that to a warning if you think that's better -- I just pushed a commit that should disable the two tests that use flags on BigQuery (those tests will fail on BQ if they don't raise).

clausherther: Looks like that still fails with the updated test https://app.circleci.com/pipelines/github/calogica/dbt-expectations/224/workflows/44eb5be9-8001-4c2e-a06c-fb6d39f31b72/jobs/136/parallel-runs/0/steps/0-106

clausherther (approved):
This is great, really well done, thanks!
Pending other PRs, I'll release this in 0.8.3 next week.

tconbeer:
Thanks, @clausherther ! Love this package

Issue #249, opened 2023-03-24
[BUG] regexp_instr does not validate regex in PostgreSQL


lookslikeitsnot:
### Is this a new bug in dbt-expectations?

- [X] I believe this is a new bug in dbt-expectations
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior
Any call to the regexp_instr macro will be valid for PostgreSQL. This is due to the generated expression and the handling of nulls by Postgres.

Dissecting the generated expression: 
```SQL
  with grouped_expression as (
    select array_length(
        (
          select regexp_matches(postal_code_5_3, '^\d{5}-\d{3}', '')
        ),
        1
      ) > 0 as expression
    from <db_name>.dbt_expectations_integration_tests.data_text
  ),

  validation_errors as (
    select *
    from grouped_expression
    where not(expression = true)
  )

  select *
  from validation_errors
```
1. `select regexp_matches('<valid_source>', '<valid_regex>', '<any_flags>')` will always return either no results or a text[]. 
2. Getting the array_length of that will either be `null` or >0
3. Comparing that to 0 will return `null` or `true` respectively
4. Comparing that to true will once again return `null` or `true` respectively
5. Negating that will return `null` or `false` which have the same full filter effect in a where clause

### Expected Behavior
regexp_instr should fail when regex doesn't match.

### Steps To Reproduce
Change any regex in "dbt-expectations/integration_tests/models/schema_tests/schema.yml" with anything else (e.g. replace the '^\\d{5}-\\d{3}' regex with 'definitely_not_a_valid_postal_code') , the tests will still succeed.

### Environment

```markdown
- OS: Manjaro Linux 22.0.2
- Python: 3.10.9
- dbt: 1.4.5
- dbt-expectations: 0.8.3
```

### Which database adapter are you using with dbt?
Postgres

### Suggested fix
Fix with the widest Postgres version compatibility range would probably be to coalesce the array_length
i.e. to replace
`array_length((select regexp_matches({{ source_value }}, '{{ regexp }}', '{{ flags }}')), 1)`
with
`coalesce(array_length((select regexp_matches({{ source_value }}, '{{ regexp }}', '{{ flags }}')), 1), 0)`
in "dbt-expectations/macros/regex/regexp_instr.sql"

clausherther:
Closed via #250

Issue #207, opened 2022-10-12
Add negative tests


clausherther:
We should add tests that are expected to fail to key tests.

PR #250, opened 2023-03-25
Fix postgres__regexp_instr not validating regex #249

lookslikeitsnot:
## What does this PR do?
Fix regex match succeeding for any pattern in Postgres (#249)
Add expected-to-fail tests for regex (#207)

## Change description
Coalesce array_length and 0 to avoid null in expression when no match exists resulting in always falsy `WHERE` condition on error check in tests.

## Type of change
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Where has this been tested?
- OS: Manjaro Linux 22.0.2
- Python: 3.10.9
- dbt: 1.4.5
- dbt-expectations: 0.8.3
- PostgreSQL: 14.6

### DIFF
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -4,18 +4,22 @@ models:
     columns:
       - name: email_address
         tests:
+            # match email address
           - dbt_expectations.expect_column_values_to_match_regex:
               regex: "@[^.]*"
+            # does not match email address, should fail
           - dbt_expectations.expect_column_values_to_match_regex:
               regex: "&[^.]*"
               config:
                 error_if: "=0"
                 warn_if: "<4"
+            # match all uppercase, but match case-insensitive (where implemented)
           - dbt_expectations.expect_column_values_to_match_regex:
               regex: "[A-Z]"
               flags: i
               config:
                 enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift' ] }}"
+            # match all uppercase, case-sensitive (where implemented), should fail
           - dbt_expectations.expect_column_values_to_match_regex:
               regex: "[A-Z]"
               flags: c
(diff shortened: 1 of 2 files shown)


clausherther:
Thanks for the PR! 
I'll look at this in more detail tomorrow, but looks like failed a test on BigQuery:
https://app.circleci.com/pipelines/github/calogica/dbt-expectations/229/workflows/09eac6bd-ccaf-46b2-af4a-e2fe299a5fad/jobs/143?invite=true#step-106-354

clausherther:
I think this (negative) test fails on BigQuery
```yaml
- dbt_expectations.expect_column_values_to_not_match_regex_list:
              regex_list: ["[A-Z]", "[0-9]"]
              flags: i
              match_on: all
              config:
                error_if: "=0"
                warn_if: "<4"
```
because this expression returns true for all rows
```sql
select 
  regexp_instr(email_address, '[A-Z]', 1, 1) = 0
  and 
  regexp_instr(email_address, '[0-9]', 1, 1) = 0
  as expression
from dbt_expectations_integration_tests.data_text
```

lookslikeitsnot:
Sorry about that, I missed the `"The flags option is not supported for BigQuery and is being ignored."` comment and forgot to add `enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift' ] }}"`.

This leads me to another question (out of scope of this ticket but might be nice to have): since BigQuery [uses re2]() and [re2 supports flags](https://github.com/google/re2/wiki/Syntax) by prepending the regex with `(?flags:re)`, could we implement that?

clausherther:
> This leads me to another question (out of scope of this ticket but might be nice to have): since BigQuery [uses re2](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#regexp_contains) and [re2 supports flags](https://github.com/google/re2/wiki/Syntax) by prepending the regex with `(?flags:re)`, could we implement that?

Sounds good! I'm not super hip re: Regex options and flags, but if that'd be useful, I'd say go ahead and open an issue for it and we'll go from there. Thanks!

clausherther (approved):

Issue #252, opened 2023-03-28
[Feature Request]Add support for regex flags for BigQuery


lookslikeitsnot:
**Is your feature request related to a problem? Please describe.**
For consistency sake since we already support flags for Postgres and Snowflake; and since the project description states it supports BigQuery, I believe it makes sense to add support for re2 flags in BigQuery regex matches .

**Describe the solution you'd like**
Although BigQuery doesn't support flags as regex function parameters, it [uses re2](https://cloud.google.com/bigquery/docs/reference/standard-sql/string_functions#regexp_contains) and [re2 supports flags](https://github.com/google/re2/wiki/Syntax) by in-lining the flags (prepending the regex with `(?flag)`). 

**Additional context**
This is a follow-up on issue #245 that added support for flags in Postgres, Snowflake, and Redshift. The idea was briefly introduced [here](https://github.com/calogica/dbt-expectations/pull/250#issuecomment-1484678785)

lookslikeitsnot:
Having flags as parameters for BigQuery instead of having them in the test pattern has a few advantages such as
1. Easier flag validation
2. Simpler adapter switching

However, it adds a few lines of code so it's going to be more work to maintain.

_Note_ In the PR, adding flags both as parameters and in the test pattern doesn't cause execution errors. The flags in the pattern take precedence over the ones in the flag (which is expected flag behavior).
e.g.: 
regex: `(?i)[A-Z]` (case-insensitive letter match)
flags: ` -i` (case-sensitive match)
will be translated to  `(?-i)(?i)[A-Z]` and match any  lower- or upper-case character.

clausherther:
Closed via #253

PR #253, opened 2023-03-28
Add support for regex flags for BigQuery

lookslikeitsnot:
## What does this PR do?
Add support for regex flags for BigQuery (#252)

## Change description
Inline flags in non-capturing block 

## Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Where has this been tested?
- OS: Manjaro Linux 22.0.2
- Python: 3.10.9
- dbt: 1.4.5
- dbt-expectations: 0.8.3
- dbt-bigquery: 1.4.3

### DIFF
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -18,7 +18,14 @@ models:
               regex: "[A-Z]"
               flags: i
               config:
-                enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift' ] }}"
+                enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift', 'bigquery' ] }}"
+            # match all uppercase with inline case-insensitive flag and case-insensitive flag parameter (where implemented)
+            # check that adapters handling flags by inlining them don't break because of the flag duplication
+          - dbt_expectations.expect_column_values_to_match_regex:
+              regex: "(?i)[A-Z]"
+              flags: i
+              config:
+                enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift', 'bigquery' ] }}"
             # match all uppercase, case-sensitive (where implemented), should fail
           - dbt_expectations.expect_column_values_to_match_regex:
               regex: "[A-Z]"
@@ -59,7 +66,7 @@ models:
               regex_list: ["[A-G]", "[H-Z]"]
               flags: i
               config:
-                enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift' ] }}"
+                enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift', 'bigquery' ] }}"
             # match all uppercase, but match case-sensitive (where implemented), should fail
           - dbt_expectations.expect_column_values_to_match_regex_list:
               regex_list: ["[A-G]", "[H-Z]"]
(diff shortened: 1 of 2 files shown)


clausherther:
Thanks! Will take a closer look tomorrow or Thursday.

lookslikeitsnot:
Test failed for Snowflake
```
Database Error in test dbt_expectations_expect_column_values_to_match_regex_data_text_email_address__i___i_A_Z_ (models/schema_tests/schema.yml)
02:05:13    100048 (2201B): Invalid regular expression: '(?i)[A-Z]', no argument for repetition operator: ?
```
I should've read the Snowflake docs more carefully, sorry. 
I'll make the test BigQuery only since its purpose is to check the in-lining of flag parameters for the BigQuery adapter.

integration_tests/models/schema_tests/schema.yml
```diff
@@ -18,7 +18,14 @@ models:
               regex: "[A-Z]"
               flags: i
               config:
-                enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift' ] }}"
+                enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift', 'bigquery' ] }}"
```
clausherther, line 21: This covers all of our support platform, I'm wondering if we should just remove the enabled config for this test then?

integration_tests/models/schema_tests/schema.yml
```diff
@@ -59,7 +66,7 @@ models:
               regex_list: ["[A-G]", "[H-Z]"]
               flags: i
               config:
-                enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift' ] }}"
+                enabled: "{{ target.type in ['postgres', 'snowflake', 'redshift', 'bigquery' ] }}"
```
clausherther, line 69: same comment here re: removed `config`

macros/regex/regexp_instr.sql
```diff
@@ -50,9 +50,31 @@ regexp_instr({{ source_value }}, '{{ regexp }}', {{ position }}, {{ occurrence }
...
+{# Flag syntax is xyz (set) or -xyz (clear) or xy-z (set xy, clear z).  #}
+
+{# Regex explanation: do not allow consecutive dashes, accept all re2 flags and clear operator, do not end with a dash  #}
+{% set re2_flags_pattern = '^(?!.*--)[-imsU]*(?<!-)$' %}
+{% set re = modules.re %}
+{% set is_match = re.match(re2_flags_pattern, flags) %}
+{% if not is_match %}
+    {# Using raise_compiler_error causes disabled tests with invalid flags to fail compilation #}
+    {{ exceptions.warn(
+        "flags " ~ flags ~ " isn't a valid re2 flag pattern"
```
clausherther, line 77: if a user passes in an unsupported flag, like `c` on BigQuery, this will generate a compiler warning, but the test will ultimately fail (e.g. "Cannot parse regular expression: invalid perl operator: (?c"). Is that the intent?

integration_tests/models/schema_tests/schema.yml
lookslikeitsnot: I've now removed all occurrences of that enabled config

macros/regex/regexp_instr.sql
lookslikeitsnot: The issue if we raise a compiler error is that the tests don't compile anymore.
As you can [see in the pipeline test](https://app.circleci.com/pipelines/github/calogica/dbt-expectations/237/workflows/e8184c91-8b40-4e48-8cc7-fef33e34c0b0/jobs/156?invite=true#step-106-9), even though the test for the "c" flag is not enabled for BigQuery (since it isn't a valid re2 flag), we have a warning in the output.
```
02:04:16  flags c isn't a valid re2 flag pattern
02:04:16  flags c isn't a valid re2 flag pattern
02:04:16  flags c isn't a valid re2 flag pattern
```
Here's the output with raise_compiler_error instead of warn:
```
Encountered an error:
Compilation Error in test dbt_expectations_expect_column_values_to_match_regex_data_text_email_address__c___A_Z_ (models/schema_tests/schema.yml)
  flags c isn't a valid re2 flag pattern
```
Having a warning is nice since the test will compile properly but it can be forced to error with the `--warn-error` flag.
i.e.:
running  `dbt --warn-error build` will fail build for invalid flags in test (even if the invalid-flag tests are disabled).
running  `dbt build` will build for invalid flags in test (but display a warning message and ultimately fail)

However, I get that this might be considered a regression since it changes the default behavior to be more lax. 
This gives us 2 options:
1. Remove the "c" flag test
2. Add support for the "c" flag in BigQuery by translating it to "-i". This improves adapter flexibility but might be unintuitive for the user since "c" is not supposed be a valid re2 flag.

clausherther: Sorry for the delayed response. I think it's to leave as is, including the c flag tests with the defined targets. If we find that this creates usability issues for folks, we can address in a later PR (and maybe implement option 2). Thanks for the thorough explanation!

clausherther (approved):

PR #259, opened 2023-05-17
Fix expect_column_most_common_value_to_be_in_set handling of ties

### LINKED ISSUE #258: [BUG] expect_column_most_common_value_to_be_in_set handling of ties

- [x] I believe this is a new bug in dbt-expectations
- [x] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

When most common values are non-unique, running expect_column_most_common_value_to_be_in_set only works for 1 of the most common values.

### Expected Behavior

expect_column_most_common_value_to_be_in_set should match all most common values

### Steps To Reproduce

in dbt-expectations/integration_tests/models/schema_tests/schema.yml, add following column test for model data_test column col_string_b.:
``` yaml
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: ['ab']
              top_n: 1
```
Although all values are distinct in that model column (i.e. all column values are "the most common"), any value in value_set different from 'b' (the first result naturally ordered) will fail.

### Relevant log output

```sql
with value_counts as (
    select
        col_string_b as value_field,
        count(*) as value_count
    from dbt_expectations_integration_tests.data_test
    group by col_string_b
),
value_counts_ranked as (
    select
        *,
        row_number() over(order by value_count desc) as value_count_rank
    from
        value_counts
),
value_count_top_n as (
    select
        value_field
    from
        value_counts_ranked
    where
        value_count_rank = 1
),
set_values as (
    select
        'ab' as value_field
),
unique_set_values as (
    select distinct value_field
    from
        set_values
),
validation_errors as (
    -- values from the model that are not in the set
    select
        value_field
    from
        value_count_top_n
    where
        value_field not in (select value_field from unique_set_values)
)
select *
from validation_errors
```
### Environment

```markdown
- OS: Manjaro Linux 22.0.2
- Python: 3.10.9
- dbt: 1.5.0
- dbt-expectations: 0.8.5
```

Postgres

- Postgres
- Snowflake
- BigQuery

### Additional Context

[Great expectations](https://greatexpectations.io/expectations/expect_column_most_common_value_to_be_in_set) has an additional ties_okay argument for partial matches. We could also add support for that.

So following tests would succeed
``` yaml
            # Expect error if not all most common values are in the set
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: ['b']
              top_n: 1
              config:
                error_if: "=0"
                warn_if: "<3"
            # Expect success if not all most common values are in the set but ties_okay is set
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: ['b']
              top_n: 1
              ties_okay: true
            # Expect error if none of the most common values are in the set and ties_okay is set
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: ['invalid_value']
              top_n: 1
              ties_okay: true
              config:
                error_if: "=0"
                warn_if: "<4"
            # Expect success if not all most common values are in the set but ties_okay is set
            # and the set contains extra values 
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: ['b', 'invalid_value']
              top_n: 1
              ties_okay: true
            # Expect success if not all most common values are in the set but ties_okay is set
            # and value is not first one of the column naturally ordered
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: ['ab']
              top_n: 1
              ties_okay: true
            # Expect success if all most common values are in the set
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: ['b', 'ab', 'abc', 'abcd']
              top_n: 1
            # Expect success if all most common values are in the set 
            # and the set contains extra values 
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: ['b', 'ab', 'abc', 'abcd', 'invalid_value']
              top_n: 1
            # Expect error if none of the most common values are in the set 
            # and the set contains extra values 
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: ['invalid_value1', 'invalid_value2', 'invalid_value3', 'invalid_value4', 'invalid_value5']
              top_n: 1
              config:
                error_if: "=0"
                warn_if: "<4"
```

lookslikeitsnot:

Closes #258
Add expected-to-fail tests for expect_column_most_common_value_to_be_in_set (https://github.com/calogica/dbt-expectations/issues/207)
  
## Summary of Changes

Fixes tie handling of expect_column_most_common_value_to_be_in_set by rank()-ing the occurrences of column values instead of row_number()-ing them.
Adds a ties_okay to validate partial matches in case of ties.

expect_column_most_common_value_to_be_in_set doesn't work reliably when multiple column values have the same occurrence. 

## Reviewers
@clausherther

### DIFF
--- a/README.md
+++ b/README.md
@@ -1009,10 +1009,12 @@ tests:
       value_set: [0.5]
       top_n: 1
       quote_values: true # (Optional. Default is 'true'.)
-      data_type: "decimal" # (Optional. Default is 'decimal')
+      data_type: "decimal"  # (Optional. Default is adapter-specific equivalent of 'decimal' with a scale provided by dbt. 
+                            # Using decimal/numeric without scale might result in unexpected behaviour with Snowflake where scale
+                            # defaults to 0 resulting in values being rounded)
       strictly: false # (Optional. Default is 'false'. Adds an 'or equal to' to the comparison operator for min/max)
-      ties_okay: true # (Optional. Default is 'false'. If true, the expectation will succeed if values outside 
-                      # the designated set are as common (but not more common) than designated values)
+      allow_ties: true # (Optional. Default is 'false'. If true, the expectation will succeed if values outside 
+                       # the designated set are as common (but not more common) than designated values)
 ```
 
 ### [expect_column_max_to_be_between](macros/schema_tests/aggregate_functions/expect_column_max_to_be_between.sql)
--- a/README.md
+++ b/README.md
@@ -1011,6 +1011,8 @@ tests:
       quote_values: true # (Optional. Default is 'true'.)
       data_type: "decimal" # (Optional. Default is 'decimal')
       strictly: false # (Optional. Default is 'false'. Adds an 'or equal to' to the comparison operator for min/max)
+      ties_okay: true # (Optional. Default is 'false'. If true, the expectation will succeed if values outside 
+                      # the designated set are as common (but not more common) than designated values)
 ```
 
 ### [expect_column_max_to_be_between](macros/schema_tests/aggregate_functions/expect_column_max_to_be_between.sql)
--- a/README.md
+++ b/README.md
@@ -1013,7 +1013,7 @@ tests:
                             # Using decimal/numeric without scale might result in unexpected behaviour with Snowflake where scale
                             # defaults to 0 resulting in values being rounded)
       strictly: false # (Optional. Default is 'false'. Adds an 'or equal to' to the comparison operator for min/max)
-      allow_ties: true # (Optional. Default is 'false'. If true, the expectation will succeed if values outside 
+      ties_okay: true # (Optional. Default is 'false'. If true, the expectation will succeed if values outside 
                        # the designated set are as common (but not more common) than designated values)
 ```
 
--- a/macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
+++ b/macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
@@ -3,13 +3,15 @@
                                                        value_set,
                                                        top_n,
                                                        quote_values=True,
-                                                       data_type="decimal",
+                                                       data_type=None,
                                                        row_condition=None,
-                                                       ties_okay=False
+                                                       allow_ties=False
                                                        ) -%}
-
+    {# For Snowflake, using a default 'decimal' instead of dbt.type_numeric() 
+        rounds up the value when casting #}
+    {% set data_type = dbt.type_numeric() if not data_type else data_type %}
     {{ adapter.dispatch('test_expect_column_most_common_value_to_be_in_set', 'dbt_expectations') (
-            model, column_name, value_set, top_n, quote_values, data_type, row_condition, ties_okay
+            model, column_name, value_set, top_n, quote_values, data_type, row_condition, allow_ties
         ) }}
 
 {%- endtest %}
@@ -21,9 +23,9 @@
                                                                       quote_values,
                                                                       data_type,
                                                                       row_condition,
-                                                                      ties_okay
+                                                                      allow_ties
                                                                       ) %}
-
+{% set data_type = data_type %}
 with value_counts as (
 
     select
@@ -93,14 +95,35 @@ most_common_values_not_in_set as (
     where
         value_field not in (select value_field from unique_set_values)
 ),
+most_common_values_in_set as (
+    select 
+        value_field 
+    from 
+        value_count_top_n 
+    {{ dbt.except() }}
+    select 
+        value_field 
+    from 
+        most_common_values_not_in_set
+),
 validation_errors as (
-    {% if ties_okay -%}
-    select mcvnis.* from most_common_values_not_in_set mcvnis
-    , (select count(*) as cnt from most_common_values_not_in_set) as most_common_values_not_in_set_cnt
-    , (select count(*) as cnt from value_count_top_n) as most_common_values_cnt
-    where most_common_values_not_in_set_cnt.cnt >= most_common_values_cnt.cnt
+    {% if allow_ties -%}
+    select 
+        * 
+    from 
+        most_common_values_not_in_set
+    where
+        {# 
+            If the intersection between the most common values and the values in the set is not empty, 
+            succeed. Otherwise fail the test and select all the most common values from the column.
+        #}
+        (
+            select count(*) 
+            from most_common_values_in_set
+        ) = 0
     {%- else -%}
-    select * from most_common_values_not_in_set
+    select * 
+    from most_common_values_not_in_set
     {%- endif -%}
 )
 
--- a/macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
+++ b/macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
@@ -5,13 +5,13 @@
                                                        quote_values=True,
                                                        data_type=None,
                                                        row_condition=None,
-                                                       allow_ties=False
+                                                       ties_okay=False
                                                        ) -%}
     {# For Snowflake, using a default 'decimal' instead of dbt.type_numeric() 
         rounds up the value when casting #}
     {% set data_type = dbt.type_numeric() if not data_type else data_type %}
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -1011,6 +1011,8 @@ tests:
       quote_values: true # (Optional. Default is 'true'.)
       data_type: "decimal" # (Optional. Default is 'decimal')
       strictly: false # (Optional. Default is 'false'. Adds an 'or equal to' to the comparison operator for min/max)
+      ties_okay: true # (Optional. Default is 'false'. If true, the expectation will succeed if values outside
```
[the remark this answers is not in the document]
lookslikeitsnot: Sure. I used ties_okay to be consistent with Great expectations' implementation
https://greatexpectations.io/expectations/expect_column_most_common_value_to_be_in_set

macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
```diff
@@ -83,15 +85,23 @@ unique_set_values as (
...
         value_field
     from
         value_count_top_n
     where
         value_field not in (select value_field from unique_set_values)
-
+),
+validation_errors as (
+    {% if ties_okay -%}
+    select mcvnis.* from most_common_values_not_in_set mcvnis
```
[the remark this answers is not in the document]
lookslikeitsnot: Is there a style guide/linter config available for this project?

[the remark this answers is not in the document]
clausherther: Ah gotcha, makes sense to leave then! I have not been keeping up with their updates ;)

macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
```diff
@@ -48,7 +52,7 @@ value_counts_ranked as (
 
     select
         *,
-        row_number() over(order by value_count desc) as value_count_rank
+        rank() over(order by value_count desc) as value_count_rank
```
clausherther, line 55: Yikes, it just occurred to me that after this, the check for top_n should be 
```sql
value_count_rank <= {{ top_n }}
```
vs
```sql
value_count_rank = {{ top_n }}
```
right? We only seem to test for `top_n: 1` so I don't think I caught this before?

I know this isn't in scope for you for this PR, but that does make sense, or am I misremembering? It's been a while since I worked on this.

clausherther: It seems to me (maybe naively) that if we make this change to `value_count_rank <= {{ top_n }}`, the change from `row_number` to `rank` should be sufficient to deal with collisions?
E.g. in `data_test` the most common value for `col_numeric_a` is `0.5` with 2 counts, with `0` and `1` both having 1 count, thus getting the same rank, `2`. 
So, this test would fail given my suggestion:
```yaml
- dbt_expectations.expect_column_most_common_value_to_be_in_set:
      value_set: [0.5, 1]
      top_n: 2
```
because `0` is just as common but not in the set, correct?

clausherther: I pushed an example to this test PR to help explain what I mean: 
https://github.com/calogica/dbt-expectations/pull/260

macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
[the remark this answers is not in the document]
clausherther: re: style guide, I tried implementing SQL Fluff, but that was not worth the effort vs the CI failures that came with it. The repo is the style guide.

README.md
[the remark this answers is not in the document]
clausherther: Ah gotcha, makes sense to keep it then! I have not been keeping up with their updates ;)

macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
lookslikeitsnot: Indeed, using `rank` and `value_count_rank <= {{ top_n }}`  solves the issue. Should I close this PR and open a feature request for "ties_okay"?

P.S.: Maybe adding following test for "col_numeric_a" might be a good idea to explicitly target top_n and prevent accidental 
```yaml
             # Expect failure if value not in column 
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: [123456789]
              quote_values: false
              top_n: >
                (select count(*) from {{ref('data_test')}})
              config:
                error_if: "=0"
                warn_if: "<>3"
```

clausherther: Feel free to modify this PR, or the other one I posted. 
I'd say we can add `ties_okay` and the change to rank and the `<=` operator all in one PR. Wdyt?

The new test makes sense (if that works with the ref in the yaml?).

Thanks!

lookslikeitsnot: Works for me! I modified this PR and added the test.

macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
```diff
@@ -83,15 +87,44 @@ unique_set_values as (
...
+most_common_values_not_in_set as (
     select
         value_field
     from
         value_count_top_n
     where
         value_field not in (select value_field from unique_set_values)
-
+),
+most_common_values_in_set as (
```
clausherther, line 98: tbh, I still don't follow why we need this whole section. Maybe I don't fully understand what we mean by ties, but I would have thought changing the ranking to `rank` from `row_number` essentially allows for ties, no?

lookslikeitsnot: Using rank will fix issues with "top_n" but I don't believe it will entirely solve the "ties_okay" issue:
what we want (to match [Great Expectations'](https://github.com/great-expectations/great_expectations/blob/develop/tests/test_definitions/column_aggregate_expectations/expect_column_most_common_value_to_be_in_set.json#L52) test) is to succeed for partial matches when ties_okay is set to true i.e.  
```yaml
models:
  - name: data_test
    columns:  
      - name: col_string_b
        tests:
          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
              value_set: ['ab', 'abc', 'abcd']
              top_n: 1
              ties_okay: true
              quote_values: false
```
Should succeed even though `b` is as common as all the other values in the data but is missing in the input. If we simply use rank, the test will fail since "most_common_values_not_in_set" only tells us whether or not **all** most common values were in the set. 
Creating "most_common_values_in_set" allows us to check if any of the input values (`['ab', 'abc', 'abcd']` in previous example) are contained in the top_n most common values (i.e. any partial match) and will succeed if there is any match (even with a missing `b`).
Maybe rewriting it as 
```sql
{# Get the partial matches for ties_okay #}
most_common_values_in_set as (
    select 
        value_field 
    from 
        value_count_top_n 
    {{ dbt.intersect() }}
    select 
        value_field 
    from 
        unique_set_values
),
```
is clearer? 
Another option is to remove most_common_values_in_set and use "having counts":
```sql
validation_errors as (
    
    select value_field 
    from most_common_values_not_in_set
    {% if ties_okay -%}
    group by 1
    {# Check that a partial match exists between the input set and the top n value 
        i.e. that the count of remaining most common values not in set is smaller 
        than the total count of most common values #}
    having not (
        (select count(*) from most_common_values_not_in_set) 
        < 
        (select count(*) from value_count_top_n)
    )
    {%- endif -%}
) 
```
This is more condensed but IMO, the `having` mixed with `select count` makes this harder to understand.
But maybe I'm missing an obvious way to get a partial match without "most_common_values_in_set"? Or maybe we don't want "ties_okay" to behave like that and only do full matches and fail for partial matches?

lookslikeitsnot:
Any reason for the closing?
No response was ever given to the comment I made 3 months ago and the issue
is still open.

PR #260, opened 2023-05-19
[Don't Merge] Issue 258 - Update to rank and fix operator

### LINKED ISSUE #258: [BUG] expect_column_most_common_value_to_be_in_set handling of ties

clausherther:

Closes #258

## Summary of Changes
Simplified example to explain suggestion made [here](https://github.com/calogica/dbt-expectations/pull/259/files#r1198452316)

## Reviewers
@clausherther

### DIFF
--- a/integration_tests/models/schema_tests/schema.yml
+++ b/integration_tests/models/schema_tests/schema.yml
@@ -505,6 +505,17 @@ models:
               value_set: [0.5]
               top_n: 1
               quote_values: false
+          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
+              value_set: [0.5, 0, 1]
+              top_n: 2
+              quote_values: false
+          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
+              value_set: [0.5, 0]
+              top_n: 2
+              quote_values: false
+              config:
+                error_if: "=0"
+                warn_if: "<>1"
           - dbt_expectations.expect_column_values_to_be_increasing:
               sort_column: col_numeric_a
               strictly: false
--- a/macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
+++ b/macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
@@ -48,7 +48,7 @@ value_counts_ranked as (
 
     select
         *,
-        row_number() over(order by value_count desc) as value_count_rank
+        rank() over(order by value_count desc) as value_count_rank
     from
         value_counts
 
@@ -60,7 +60,7 @@ value_count_top_n as (
     from
         value_counts_ranked
     where
-        value_count_rank = {{ top_n }}
+        value_count_rank <= {{ top_n }}
 
 ),
 set_values as (


integration_tests/models/schema_tests/schema.yml
```diff
@@ -505,6 +505,20 @@ models:
               value_set: [0.5]
               top_n: 1
               quote_values: false
+              data_type: float
+          - dbt_expectations.expect_column_most_common_value_to_be_in_set:
+              value_set: [0.5, 0, 1]
+              top_n: 2
+              quote_values: false
+              data_type: float
```
clausherther, line 513: fyi, Snowflake won't correctly compare a `float` in the db to a value cast to `decimal` (default for the test)

macros/schema_tests/aggregate_functions/expect_column_most_common_value_to_be_in_set.sql
```diff
@@ -22,6 +22,8 @@
                                                                       row_condition
                                                                       ) %}
 
+{% set data_type = dbt.type_numeric() if not data_type else data_type %}
```
clausherther, line 25: fyi, Snowflake won't correctly compare a float in the db to a value cast to decimal (default for the test) so makes more sense to use the built-in types across the test to make sure comparisons actually work.

lookslikeitsnot: From [the official documentation](https://docs.snowflake.com/en/sql-reference/data-types-numeric) and my testing,  inserting floating point data will create a `number` column type with the smallest precision and scale to fit the value. e.g.:  when creating a table with initial data, if a column contains [0, 1, 0.5], the assigned type will be `number(2,1)`. So no information is lost at that point. However casting that to "numeric", "decimal", "number" type without explicit precision and scale will result in rounding and casting it to any "float" or "double" type [has the usual accuracy problems](https://en.wikipedia.org/wiki/Floating-point_arithmetic#Accuracy_problems).

When comparing, Snowflake will happily attempt to compare any numeric type with the usual issues from floats
 <table>
  <tr>
    <th>comparison</th>
    <th>result</th>
  </tr>
  <tr>
   <td>cast(0.5 as number(2,1)) = cast(0.5 as numeric(28,2))</td>
   <td>TRUE</td>
     </tr>
  <tr>
   <td>cast(0.5 as float) = cast(0.5 as numeric(28,2))</td>
   <td>TRUE</td>
     </tr>
  <tr>
    <td>cast(0.5 as float) = cast(0.5 as numeric)</td>
    <td>FALSE</td>
     </tr>
<tr>
    <td>0.1 + 0.2 = 0.3</td>
    <td>TRUE</td>
  </tr>
<tr>
    <td>(cast(0.1 as float) + cast(0.2 as float)) = cast(0.3 as float)</td>
    <td>FALSE</td>
  </tr>
  <tr>
    <td>cast(3.1415926535897932384626433832795028841 as number(38,37)) = cast(3.141592653589793 as float)</td>
    <td>TRUE</td>
  </tr>
</table> 

*TL;DR*: It makes sense to use built-in types everywhere to not get db-engine specific issues.

integration_tests/models/schema_tests/data_test.sql
```diff
@@ -13,8 +13,8 @@ union all
 select
     2 as idx,
     '2020-10-22' as date_col,
-    1 as col_numeric_a,
-    0 as col_numeric_b,
+    cast(1 as {{ dbt.type_numeric() }}) as col_numeric_a,
```
lookslikeitsnot, line 16: Is there any database engine that wouldn't automatically cast every subsequent value on insert when the first value has an explicit type? (it doesn't hurt to cast them all, I'm curious and couldn't find the information online)

PR #263, opened 2023-06-05
Add a new macro to test numeric types with specific precision and scale

samousavih:

Closes #(262) 

If you don't have an issue #, please first open an issue on the repo before submitting a PR to discuss the changes you'd like to make.
  
## Summary of Changes

1) A new macro to assert numeric types and their scale and precision

Current macro to assert column types doesn't include scale and precision.

  
## Reviewers
@clausherther


clausherther:
Hi, sorry for the delayed response! Instead of making a new test, would it be possible to make this an optional path in `expect_column_values_to_be_of_type`?

samousavih:
Hey @clausherthe, thanks for the reply,
One option is to make 'expect_column_values_to_be_of_type' and 'expect_column_values_to_be_in_type_list' assert precision and scale when mentioned and ignore them when not, which is potentially what most would expect from the test. 
```
- dbt_expectations.expect_column_values_to_be_of_type:
      column_type: numeric(2,1)
- dbt_expectations.expect_column_values_to_be_in_type_list:
      column_type_list: [numeric(2,1)]
```
```
- dbt_expectations.expect_column_values_to_be_of_type:
      column_type: numeric
- dbt_expectations.expect_column_values_to_be_in_type_list:
      column_type_list: [numeric]
```
How does this look like to you?
Only minor issue might be it could fail existing tests once package users upgrade, which could be a good thing depending how you look at it.

PR #273, opened 2023-08-07
add basic column tests for arrays and update readme

ilkernator:

#272

  
## Summary of Changes

Added two new tests to the "column_values_basic" directory and updated the readme-file:
- expect_array_column_to_not_be_empty
- expect_struct_column_to_not_be_empty

For tables that contain columns of type `array` or `array(struct)`, the values can be sometimes empty without being NULL (e.g. `[ ]`, or `[" "]`. In this case, the test to **expect_column_values_to_not_be_null.sql** could be misleading and mask the issue of having empty values. So we would need to make use of Hive functions to assert the array is indeed empty.

  
## Reviewers
@clausherther

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ For example, use `America/New_York` for East Coast Time.
 - [expect_column_values_to_be_of_type](#expect_column_values_to_be_of_type)
 - [expect_column_values_to_be_in_type_list](#expect_column_values_to_be_in_type_list)
 - [expect_column_values_to_have_consistent_casing](#expect_column_values_to_have_consistent_casing)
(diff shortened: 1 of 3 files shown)


clausherther:
Hi, thanks for this PR. 
Before I can take a look at this, we'll need to add integration tests for both macros, which requires adding the appropriately typed columns to the test datasets.
Also, I think I'd prefer implementing this more generically as a single test, ~ `expect_column_to_not_be_empty`. Would that be possible across all platforms we support? 
Also, we'd need the positive equivalent, `expect_array_column_to_be_empty`.

PR #291, opened 2023-11-02
Fix sum_total in expect_multicolumn_sum_to_equal to accept column names

### LINKED ISSUE #290: [BUG] expect_multicolumn_sum_to_equal fails if sum_total is another column

- [x ] I believe this is a new bug in dbt-expectations
- [x ] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior
expect_multicolumn_sum_to_equal fails if sum_total is another column

if I have a source table like
```
      - name: test_counts
        tests:
          - dbt_expectations.expect_multicolumn_sum_to_equal:
              column_list: ["test_pass_count", "test_fail_count", "test_error_count", test_abort_count]
              sum_total: test_count #I can workaround by just doing sum(test_count)
```

### Expected Behavior
I would expect it to not break

### Steps To Reproduce
just put a different column in `sum_total`
### Relevant log output
 [MISSING_GROUP_BY] The query does not include a GROUP BY clause. Add GROUP BY or turn it into the window functions using OVER clauses.
### Environment

```markdown
- OS: windows 11
- Python: 3.9
- dbt: 1.6.6, dbt-databricks:1.6.6
- dbt-expectations: 0.8.5
```

- Postgres
- Snowflake
- BigQuery

VDFaller:

Closes #290 
  
## Summary of Changes

Just wrapped `sum_total` in Sum() 

This will allow a column name or a literal number to be used.  

  
## Reviewers
@clausherther

### DIFF
--- a/macros/schema_tests/multi-column/expect_multicolumn_sum_to_equal.sql
+++ b/macros/schema_tests/multi-column/expect_multicolumn_sum_to_equal.sql
@@ -9,7 +9,7 @@
 {% set expression %}
 {% for column in column_list %}
 sum({{ column }}){% if not loop.last %} + {% endif %}
-{% endfor %} = {{ sum_total }}
+{% endfor %} = sum({{ sum_total }})
 {% endset %}
 
 {{ dbt_expectations.expression_is_true(model,
--- a/macros/schema_tests/multi-column/expect_multicolumn_sum_to_equal.sql
+++ b/macros/schema_tests/multi-column/expect_multicolumn_sum_to_equal.sql
@@ -9,7 +9,8 @@
 {% set expression %}
 {% for column in column_list %}
 sum({{ column }}){% if not loop.last %} + {% endif %}
-{% endfor %} = sum({{ sum_total }})
+{# the if just allows for column names or literal numbers #}
+{% endfor %} = {% if sum_total is number %}{{sum_total}}{% else %}sum({{ sum_total }}){% endif %}
 {% endset %}
 
 {{ dbt_expectations.expression_is_true(model,
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,7 +119,7 @@ jobs:
           command: |
             python3 -m venv venv
             . venv/bin/activate
-            pip install dbt-spark "dbt-spark[PyHive]"
+            pip install "dbt-spark==$DBT_VERSION" "dbt-spark[PyHive]==$DBT_VERSION"
       - run: *dbt-deps
       - run:
           name: Wait for Spark-Thrift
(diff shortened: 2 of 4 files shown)


clausherther:
Great catch, thanks for the PR! I think this is one of those holdovers from Great Expectations where tests are built against static values vs totals of other columns. Probably worth looking other tests as well.

Would you mind also adding a test that covers this new use case of testing against the sum of another column? We currently only have a static test, i.e.
```yaml
      - dbt_expectations.expect_multicolumn_sum_to_equal:
          column_list: ["col_numeric_a", "col_numeric_b"]
          sum_total: 4
```
so we'll need something like 
```yaml
      - dbt_expectations.expect_multicolumn_sum_to_equal:
          column_list: ["col_numeric_a", "col_numeric_b"]
          sum_total: new_totals_column
```

VDFaller:
Got most of the tests made. One snag, sorry haven't used postgres in a while.  Can you tell me why 

```
SELECT Sum(4)
from postgres.dbt_expectations_integration_tests.data_test
``` 
returns 16
but 
```
SELECT sum(4)
```
RETURNS 4?  

Am I just insane that something is wrong?  

Postgres version: 
PostgreSQL 14.1 on x86_64-pc-linux-musl, compiled by gcc (Alpine 10.3.1_git20211027) 10.3.1 20211027, 64-bit

VDFaller:
Hopefully this is what you're looking for.  If not let me know.

clausherther:
> Hopefully this is what you're looking for. If not let me know.

I'll check it out! FYI, you probably figured this out since, but `sum(4)` against the table return `16` b/c it does this for every one of the 4 rows. This may actually mean this is a break change. Any config that checks for 4 would now fail. Any thoughts?

macros/schema_tests/multi-column/expect_multicolumn_sum_to_equal.sql
```diff
@@ -9,7 +9,8 @@
 {% set expression %}
 {% for column in column_list %}
 sum({{ column }}){% if not loop.last %} + {% endif %}
-{% endfor %} = {{ sum_total }}
+{# the if just allows for column names or literal numbers #}
+{% endfor %} = {% if sum_total is number %}{{sum_total}}{% else %}sum({{ sum_total }}){% endif %}
```
clausherther, line 13: 👍

clausherther:
FYI, updated the CI config on this PR to get Spark CI to pass. For some reason our CI configs breaks in dbt 1.7.x, but I'll fix that separately. CI is passing for you now 👍

clausherther (approved):
👍

PR #294, opened 2023-11-07
Add Trino Support

### LINKED ISSUE #293: Trino support
Add Trino support

damian3031:

Closes #293 
  
## Summary of Changes

- Added Trino support
- Fixed expect_select_column_values_to_be_unique_within_record macro
- Fixed timeseries_data_extended.row_value expect_column_distinct_count_ tests row_conditions
  
## Reviewers
@clausherther

### DIFF
--- a/macros/schema_tests/aggregate_functions/expect_column_proportion_of_unique_values_to_be_between.sql
+++ b/macros/schema_tests/aggregate_functions/expect_column_proportion_of_unique_values_to_be_between.sql
@@ -6,7 +6,7 @@
                                                             strictly=False
                                                             ) %}
 {% set expression %}
-count(distinct {{ column_name }})*1.0/count({{ column_name }})
+count(distinct {{ column_name }})*1.000/count({{ column_name }})
 {% endset %}
 {{ dbt_expectations.expression_between(model,
                                         expression=expression,
--- a/macros/schema_tests/aggregate_functions/expect_column_proportion_of_unique_values_to_be_between.sql
+++ b/macros/schema_tests/aggregate_functions/expect_column_proportion_of_unique_values_to_be_between.sql
@@ -6,7 +6,7 @@
                                                             strictly=False
                                                             ) %}
 {% set expression %}
-count(distinct {{ column_name }})*1.000/count({{ column_name }})
+cast(count(distinct {{ column_name }}) as {{ dbt.type_float() }})/count({{ column_name }})
 {% endset %}
 {{ dbt_expectations.expression_between(model,
                                         expression=expression,
--- a/macros/schema_tests/multi-column/expect_select_column_values_to_be_unique_within_record.sql
+++ b/macros/schema_tests/multi-column/expect_select_column_values_to_be_unique_within_record.sql
@@ -40,7 +40,7 @@
 with column_values as (
 
     select
-        row_number() over(order by 1) as row_index,
+        row_number() over(order by {{ columns|join(', ') }}) as row_index,
         {% for column in columns -%}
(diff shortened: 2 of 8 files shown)


damian3031:
@clausherther Could you approve the workflow in CircleCI, to verify if all tests will pass?

clausherther:
Thanks for this! I'll review by end of week.

macros/schema_tests/multi-column/expect_select_column_values_to_be_unique_within_record.sql
```diff
@@ -40,7 +40,7 @@
 with column_values as (
 
     select
-        row_number() over(order by 1) as row_index,
+        row_number() over(order by {{ columns|join(', ') }}) as row_index,
```
clausherther, line 43: Was this failing? Is this required for Trino support? I'd have to look at this separately to step through, it's been a while, but if this is a bug, I'd rather roll this into a separate PR.

macros/schema_tests/aggregate_functions/expect_column_proportion_of_unique_values_to_be_between.sql
```diff
@@ -6,7 +6,7 @@
                                                             strictly=False
                                                             ) %}
 {% set expression %}
-count(distinct {{ column_name }})*1.0/count({{ column_name }})
+count(distinct {{ column_name }})*1.000/count({{ column_name }})
```
clausherther, line 9: Does Trino not do integer division with `1.0`?

damian3031: In Trino, literal `1.0` has type `decimal(2,1)`. When multiplying integers with it, the result is of decimal type, with the same scale. Therefore, any results are rounded to 1 decimal point in this case. Because of that, [this test](https://github.com/calogica/dbt-expectations/blob/0.10.1/integration_tests/models/schema_tests/schema.yml#L526) is failing, as value with 2 decimal points is specified in the test. Using 3 decimal points should be sufficient for most use cases, I think.

Maybe it would be a cleaner solution to explicitly cast as a float, instead of multiplying by 1.0?
```suggestion
cast(count(distinct {{ column_name }}) as {{ dbt.type_float() }})/count({{ column_name }})
```

macros/schema_tests/multi-column/expect_select_column_values_to_be_unique_within_record.sql
damian3031: In Trino, ordering by column's position number inside ranking functions doesn't work. `row_number() over(order by 1)` results in a random order. [This test](https://github.com/calogica/dbt-expectations/blob/0.10.1/integration_tests/models/schema_tests/schema.yml#L474-L476) fails because of that.

However, IMHO it is a bug anyway. 

Mentioned test passes in other dbt adapters, because in [current test model](https://github.com/calogica/dbt-expectations/blob/0.10.1/integration_tests/models/schema_tests/data_test.sql) the first column `idx` contains subsequent id numbers. But, for example, we could have same value for all rows in the first column, and we will receive a random order.

IMHO ordering by all columns that are expected to have unique values makes more sense, than ordering by the first arbitrary column.

WDYT? I can create a separate PR for this change, if you agree with the above.

wjhrdy:
Thanks for all the hard work on this.

I really appreciate it.

damian3031:
Hi @clausherther, any update here?

clausherther (approved):
Sorry for the long wait on this one!

PR #303, opened 2024-03-30
Update README.md

thomaslaber:
Simply increased the version numbers in the snippet for how to add the dependency the `packages.yml`. This seems to have been forgotten when version 10 was realsed. At least, that is how you handled it the last few releases. 

Closes #(Issue Number) 
  
## Summary of Changes

I do not think there is an issues or a discussion needed. Maybe there is a reason why you didn't update but my guess is it was simply forgotten. 

Correct snippet for the end user to copy. 

  
## Reviewers
@clausherther

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Include in `packages.yml`
 ```yaml
 packages:
   - package: calogica/dbt_expectations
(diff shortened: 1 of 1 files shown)


clausherther (approved):
Good catch, thanks!

PR #316, opened 2024-08-14
Adds optional `step` param to expect_column_values_to_be_{in,de}creasing

### LINKED ISSUE #315: [Feature Request] Allow defining the step between values on increasing/decreasing tests

I have a column that must be strictly increasing, but the values need to have a difference of exactly 1 between them. So values like 1, 2, 3 are valid, but 1, 3, 4 are not.

It would be great if `expect_column_values_to_be_increasing` accepted an optional `step` parameter. If this is set, then it requires increasing values with this specific step. Otherwise, the test continue working as now.

I couldn't find a test that would work like this.

**Additional context**

N/A

vitorbaptista:
If set, the test will check that the values are not only strictly increasing/decreasing, but the difference between subsequent values is exactly `step`.

Closes #315
  
## Summary of Changes

Adds optional field `step` to tests `expect_column_values_to_be_{in,de}creasing`. If this value is set, it requires that the difference between subsequent values is exactly `step`. If it's not set, the behaviour of the test is the same as now (i.e., this isn't a breaking change).

The caveat is that it will only work for numeric columns. The test is literally `value_field - value_field_lag = step`.

I wrote this test for my own project. As I've written it already, I decided to sent this PR even without discussions. I'm able to do any changes required to get it merged. It's also OK if you don't think this is a good improvement to the project.

I have a column that must be strictly increasing, but the values need to have a difference of exactly 1 between them. So values like 1, 2, 3 are valid, but 1, 3, 4 are not.

## Reviewers
@clausherther

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,6 +566,7 @@ tests:
       row_condition: "id is not null" # (Optional)
       strictly: true # (Optional for comparison operator. Default is 'true', and it uses '>'. If set to 'false' it uses '>='.)
       group_by: [group_id, other_group_id, ...] # (Optional)
+      step: 1 # (Optional. If set, it requires the difference between values to be exactly this step. Requires numeric columns.)
 ```
 
 ### [expect_column_values_to_be_decreasing](macros/schema_tests/column_values_basic/expect_column_values_to_be_decreasing.sql)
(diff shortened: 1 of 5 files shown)


clausherther:
Can we add a test to https://github.com/calogica/dbt-expectations/blob/main/integration_tests/models/schema_tests/schema.yml that implements `step`?

vitorbaptista:
@clausherther Sure! It's done.

clausherther:
@vitorbaptista CI is failing b/c the macros also need to have `step` in the signature, with a default of `None`, e.g. 
```
{% test expect_column_values_to_be_increasing(model, column_name,
                                                   sort_column=None,
                                                   strictly=True,
                                                   row_condition=None,
                                                   group_by=None,
                                                   step=None
) %}
```

vitorbaptista:
Ops! I wrote that change locally but didn't commit it. Pushing the fix now.

vitorbaptista:
@clausherther just letting you know that AFAIK this is good to go :v:

clausherther (approved):

Issue #321, opened 2024-11-01
[Feature Request] `expect_table_row_count_to_equal` should return the actual row count as well as the expression


siljamardla:
I want to monitor some row counts on my table. I'm planning to set up row count tests and enable storing failures. 
I plan to do this with multiple different conditions. I actually expect some of these two fail. The aim is to monitor how **badly** they fail. The idea is to visualise the outcome in a calendar heat map, to give me a visual on when the issues are big. 

I could use the row count test, configured like this:
```yml
      - dbt_expectations.expect_table_row_count_to_equal:
          value: 0
          group_by: [created_date]
          row_condition: >
            some_column='failed'
          config:
            severity: warn
            store_failures: true
```

However, the outcome is a true/false only. 

Sample compiled code:
```sql
    with grouped_expression as (
    select
        created_date as col_1,
        count(*) = 0 as expression
    from 
      schema_name.table_name
    where
       some_column='failed'
    group by  created_date
),
validation_errors as (
    select *
    from
        grouped_expression
    where
        not(expression = true)
)
select *
from validation_errors
```

<img width="307" alt="image" src="https://github.com/user-attachments/assets/e3dfe6c8-5fc3-4dc5-a310-3f28b3bb7674">

I wish the compiled code would have an additional column for the actual row_count, like this:
Sample compiled code:
```sql
    with grouped_expression as (
    select
        created_date as col_1,
        count(*) as row_count,
        count(*) = 0 as expression
    from 
      schema_name.table_name
    where
       some_column='failed'
    group by  created_date
),
validation_errors as (
    select *
    from
        grouped_expression
    where
        not(expression = true)
)
select *
from validation_errors
```
to return 
<img width="443" alt="image" src="https://github.com/user-attachments/assets/f41607ba-57ec-4631-8032-8aac1e1b13e6">

It's not a breaking change as the logic of the test remains the same. We're just making the stored output more informative. 

I'm happy to contribute, but I could not immediately figure out where to make this change. The actual code compiling happens through so many macros that I'm not (yet) familiar with.

siljamardla:
I did figure out where to do the change. I've made a specific change to this test, also duplicating 2 generic macros to achieve my goal. Clearly this is not the way to solve the question in long term. I'm happy to work together to make it a cleaner and more uniform update, just don't want to go down a rabbit hole without any confirmation from maintainers' side :)

PR #323, opened 2024-11-07
sample changes to expose expression LHS and RHS

siljamardla:

Shows how to address #321 

## Summary of Changes

In the PR I've shown how to expose the left and right hand side of the expression in the final test result, but I've only done so in one test. The same macros that I have modified are also used in many other tests that would be affected. 

To make the test outcome more useful. 

image
  
## Reviewers
@clausherther


clausherther:
FYI, closing this since we'll be stopping support for this package by the end of year.

PR #325, opened 2024-11-29
Adding support for ClickHouse specific lagInFrame function in tests

### LINKED ISSUE #324: [Feature Request] ClickHouse support in tests using lag function
ClickHouse doesn't have the standard _lag_ function, instead it has _lagInFrame_ function so tests using _lag_ fail

I have already implemented the suggested workaround from the ClickHouse [docs](https://clickhouse.com/docs/en/sql-reference/window-functions#:~:text=lag/lead(value%2C%20offset)) and would like to push it into the main repo so that others can use it as well

Implementing the solution

**Additional context**
n/a

zstipanicev:

Closes #324 
  
## Summary of Changes

New utility macro _lag_ was added to support ClickHouse specific _lagInFrame_ function and to ensure support for all other sql dialects while keeping the test code clean and readable. This macro is used in the 3 tests which use _lag_ function to ensure ClickHouse compatibility

It's needed to add support for ClickHouse in the current 3 tests which use the _lag_ function and any future tests which will use _lag_ function.

  
## Reviewers
@clausherther

### DIFF
--- a/macros/schema_tests/column_values_basic/expect_column_values_to_be_decreasing.sql
+++ b/macros/schema_tests/column_values_basic/expect_column_values_to_be_decreasing.sql
@@ -7,6 +7,7 @@
 
 {%- set sort_column = column_name if not sort_column else sort_column -%}
 {%- set operator = "<" if strictly else "<=" %}
+{%- set partition_by = group_by | join(", ") if group_by else "" -%}
 with all_values as (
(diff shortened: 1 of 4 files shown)


clausherther:
Hi, we don't support ClickHouse since we do not have CI in place to test this.

PR #326, opened 2024-12-03
Enable overriding lag function

### LINKED ISSUE #324: [Feature Request] ClickHouse support in tests using lag function

zstipanicev:

Closes #324 
  
## Summary of Changes

New utility macro _lag_ was added to enable overriding _lag_ function for dialects which do not support stand version of the function (e.g. ClickHouse). Also. two of the tests, using lag, which did not support dispatch were refactored to support it to enable overriding.
This PR doesn't contain any ClickHouse specific code.

It's needed to enable using tests which utilize _lag_ function on projects using sql dialects which do not support standard _lag_ function by overriding lag macro.

  
## Reviewers
@clausherther

### DIFF
--- a/macros/schema_tests/column_values_basic/expect_column_values_to_be_decreasing.sql
+++ b/macros/schema_tests/column_values_basic/expect_column_values_to_be_decreasing.sql
@@ -7,6 +7,7 @@
 
 {%- set sort_column = column_name if not sort_column else sort_column -%}
 {%- set operator = "<" if strictly else "<=" %}
+{%- set partition_by = group_by | join(", ") if group_by else "" -%}
 with all_values as (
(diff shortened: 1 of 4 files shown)


clausherther:
FYI, closing this since we'll be stopping support for this package by the end of year.

PR #328, opened 2025-06-05
Drop deprecated sqlfluff recurse config

sambradbury:

Closes #(Issue Number) 

If you don't have an issue #, please first open an issue on the repo before submitting a PR to discuss the changes you'd like to make.
  
## Summary of Changes

Removes a config option deprecated by [sqlfluff here](https://docs.sqlfluff.com/en/stable/reference/releasenotes.html). 

Removing this option will prevent users from running into errors like the one below while trying to use sqlfluff in a dbt project where dbt-expectations is installed. 
```
==== finding fixable violations ====

User Error: Config file PosixPath('/Users/myusername/Documents/dbt_project/dbt_packages/dbt_expectations/integration_tests/.sqlfluff') set an outdated config value core:recurse.

Removed as unused in production and unnecessary for debugging.

See https://docs.sqlfluff.com/en/stable/perma/configuration.html for more details.
```

  
## Reviewers
@clausherther


sambradbury:
Closing because I realize this is better placed in: https://github.com/metaplane/dbt-expectations

